PR #26, opened 2014-11-19
making ouija use config.py for user with docker

jamonation:
This pull request is to make ouija work with Docker, which I have built over in https://github.com/jamonation/ouija_docker (based on https://bugzilla.mozilla.org/show_bug.cgi?id=1010465)

This repository is a git submodule inside the docker container so that it can remain as a standalone python virtualenv, in addition to working inside Docker.

Testing and feedback are much appreciated!

### DIFF
--- a/config/config.py
+++ b/config/config.py
@@ -1,0 +1,3 @@
+DB_USER = 'ouija_user'
+DB_HOST = 'database'
+DB_NAME = 'ouija'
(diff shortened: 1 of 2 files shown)


jmaher:
I am going to close this, a lot has changed with ouija over the last 2 years- possibly we can revisit this in the future if needed.

PR #31, opened 2015-01-13
Adding single entries for a day to database and having a diff. option. r...

vaibhavmagarwal:
...=jmaher

### DIFF
--- a/tools/failures.py
+++ b/tools/failures.py
@@ -3,269 +3,286 @@ import re
 import copy
 import httplib
 import datetime
-
-def removeType(master, targetlist):
-    """
-        Given a master dictionary of:
-           key: failure id (unique bugid comment representing revision)
-           value: array of instance tuples that meet this requirement
-
-        iterate through each key and remove matching targetlist items 
-    """
-    if targetlist == []:
-        return master
-
-    removed = {}
-    for row in master:
-        if targetlist[0] in master[row]:
-            if row not in removed:
-                removed[row] = []
-            removed[row].append(targetlist[0])
-
-
-    for row in removed:
-        for item in removed[row]:
-            master[row].remove(item)
-        if len(master[row]) == 0:
-            del master[row]
-
-    return master, removed
-
-def addType(master, toAdd):
-    for row in toAdd:
-        if row not in master:
-            master[row] = []
-
-        master[row].extend(toAdd[row])
-
-    return master
-
-def bruteForcePercentage(total, scenarios, master, results, levels):
-    """
-        This is one level deep for brute force calculations.
-
-        We take an array of scenarios (in tuple format), and for each one
-        we test a removal of this specific scenario and record the scenario
-        in a dictionary of results with this format:
-           Key: percent match (i.e 99)
-           Value: list of tuple sets which can be removed and allow us to gain [Key] coverage
-    """
-    for scenario in scenarios:
-        temp, removed = removeType(master, scenario)
-
-        p = int((float(len(temp)) / total) * 100)
-        if p not in results:
-            results[p] = []
-        results[p].append(levels + scenario)
-        master = addType(master, removed)
-    return results
-    
-
-def removeBadScenarios(results, threshold):
-    """
-        Given a threshold, remove anything from a results dictionary
-        that is below the threshold.  Results dictionary format is:
-            Key: percent match (i.e 99)
-            Value: list of tuple sets which can be removed and allow us to gain [Key] coverage
-
-        This is used as an optimization, so we focus on important scenarios
-
-        Return value is a Results dictionary
-    """
-    retVal = {}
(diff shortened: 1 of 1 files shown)


tools/failures.py
```diff
@@ -195,27 +197,26 @@ def depth_first(jobtype, target):
...
-    tuples_existing = run_query("select tuple from seta");
-    tuples_to_enable = compare_array(tuples_existing, to_remove, 'delete from seta where tuple="%s"')
-
     now = datetime.datetime.now()
-    query = 'insert into seta (date, tuple) values ("%s", ' % now.strftime('%Y-%m-%d 00:00:00')
-    query += '"%s")'
-    tuples_to_disable = compare_array(to_remove, tuples_existing, query)
+    run_query('delete from seta where date="%s"' % now.strftime('%Y-%m-%d 00:00:00'))
+    for tuple in to_remove:
+        query = 'insert into seta (date, tuple) values ("%s", ' % now.strftime('%Y-%m-%d 00:00:00')
```
jmaher: I would cache the current date instead of doing now.strftime(...) twice.

tools/failures.py
```diff
@@ -195,27 +197,26 @@ def depth_first(jobtype, target):
...
 
     format_in_table(platforms, buildtypes, testtypes, to_remove)
     print "We will detect %.2f%% (%s) of the %s failures" % (percent_detected, len(total_detected), total)
 #    print "Total Time: %.2f minutes" % (total_time / 60)
 #    print "Saved Time: %.2f minutes (%.2f%%)" % ((saved_time / 60), ((saved_time / total_time)*100))
 #    print "New Time: %.2f minutes" % ((total_time - saved_time) / 60)
-    print "we should turn on: %s" % tuples_to_enable
-    print "we should turn off: %s" % tuples_to_disable
+#    print "we should turn on: %s" % tuples_to_enable
+#    print "we should turn off: %s" % tuples_to_disable
```
jmaher: remove these comments.

tools/failures.py
```diff
@@ -229,9 +230,63 @@ def run_query(query):
...
+def check_data(query_date):
+    data = run_query('select tuple from seta where date="%s"' % query_date)
+    if not data:
+        print "The database does not have data for the given %s date." % query_date
+        for date in range(-3, 4):
+            current_date = query_date + datetime.timedelta(date)
+            tuple = run_query('select tuple from seta where date="%s"' % current_date)
+            if tuple:
+                print "The data is available for date=%s" % current_date
+        return None
```
jmaher: I think we could simplify this- possibly a future bug/improvement.  I am thinking something like:
select distinct date from seta where date>old_date and date <new_date

tools/failures.py
```diff
@@ -229,9 +230,63 @@ def run_query(query):
     cur.close()
     return results
 
+def check_data(query_date):
+    data = run_query('select tuple from seta where date="%s"' % query_date)
+    if not data:
+        print "The database does not have data for the given %s date." % query_date
+        for date in range(-3, 4):
+            current_date = query_date + datetime.timedelta(date)
+            tuple = run_query('select tuple from seta where date="%s"' % current_date)
```
jmaher: s/tuple/jobtype/ <- tuple is sort of confusing, so the table column is now called jobtype!

tools/failures.py
```diff
@@ -3,269 +3,288 @@
...
+    end_tuple = check_data(options.end_date)
+
+    if start_tuple is None or end_tuple is None:
+        return
+    else:
+        print "The tuples present on %s and not present on %s: " % (options.end_date, options.start_date)
+        result = list(set(start_tuple) - set(end_tuple))
+        print result
+        if len(result) > 20:
+            print "Warning: Too many differences detected, most likely there is data missing or incorrect"
```
jmaher: I would do a if else here:
if len(result) > 20:
  print warning
else:
  print result

likewise for the line below.

tools/failures.py
```diff
@@ -3,269 +3,288 @@
...
+    for tuple in to_remove:
+        query = 'insert into seta (date, jobtype) values ("%s", ' % now
+        query += '"%s")' % tuple
+        run_query(query)
+
+    format_in_table(platforms, buildtypes, testtypes, to_remove)
+    print "We will detect %.2f%% (%s) of the %s failures" % (percent_detected, len(total_detected), total)
+#    print "Total Time: %.2f minutes" % (total_time / 60)
+#    print "Saved Time: %.2f minutes (%.2f%%)" % ((saved_time / 60), ((saved_time / total_time)*100))
+#    print "New Time: %.2f minutes" % ((total_time - saved_time) / 60)
```
jmaher: feel free to remove these comments

jmaher:
this is a great start, and should give us a foundation to build off of :)

PR #33, opened 2015-01-14
Adding an option for sanity check via the --quick option for speeding up...

vaibhavmagarwal:
... the process. r=jmaher


jmaher:
thanks, this looks much better

PR #36, opened 2015-01-15
add seta data api endpoints

jmaher:
for seta to be successful we need to have a way to get reports on demand, this basic api gives us the data.  next up will be a web page to display this.

### DIFF
--- a/tools/failures.py
+++ b/tools/failures.py
@@ -9,9 +9,11 @@ import sys
 from argparse import ArgumentParser
 from emails import send_email
 
-def getRawData(jobtype):
+import seta
+
+def getRawData():
     conn = httplib.HTTPConnection('alertmanager.allizom.org')
-    cset = "/data/seta/"
+    cset = "/data/seta/?startDate=2014-10-16&endDate=2015-01-16"
     conn.request("GET", cset)
     response = conn.getresponse()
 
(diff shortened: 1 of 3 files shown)


tools/failures.py
```diff
@@ -9,9 +9,11 @@
 from argparse import ArgumentParser
 from emails import send_email
 
-def getRawData(jobtype):
+import seta
+
+def getRawData():
     conn = httplib.HTTPConnection('alertmanager.allizom.org')
-    cset = "/data/seta/"
+    cset = "/data/seta/?startDate=2014-10-16&endDate=2015-01-16"
```
dminor: Why are these dates hard coded?

tools/failures.py
```diff
@@ -234,23 +122,18 @@ def sanity_check(jobtype, target, options):
...
+        return to_remove, percent_detected
     else:
-        return -1
+        return [], total_detected
 
 def run_query(query):
     db = MySQLdb.connect(host="localhost",
                          user="root",
-                         passwd="root",
+                         passwd="testing",
```
dminor: I think this still has to be 'root'

tools/seta.py
```diff
@@ -0,0 +1,111 @@
...
+                 'mochitest-other', 'reftest', 'xpcshell', 'crashtest', 'jsreftest', 'reftest-ipc', 'crashtest-ipc',
+                 'mochitest-browser-chrome-1', 'mochitest-browser-chrome-2', 'mochitest-browser-chrome-3',
+                 'mochitest-devtools-chrome-1', 'mochitest-devtools-chrome-2', 'mochitest-devtools-chrome-3', 'mochitest-devtools-chrome']
+    buildtypes = ['debug', 'opt']
+
+    return platforms, buildtypes, testtypes
+
+def is_matched(f, removals):
+    found = False
+    for tuple in removals:
```
dminor: I think you should use a more descriptive variable name than 'tuple'

dminor:
Sorry for the delay in looking at this, I was on PTO and then travelling.

PR #38, opened 2015-01-18
Adding more platforms to seta. r=jmaher

vaibhavmagarwal:
After adding these platforms, the output looks like this - https://pastebin.mozilla.org/8243481

### DIFF
--- a/tools/emails.py
+++ b/tools/emails.py
@@ -11,8 +11,8 @@ def send_email(regression, jobs, message, addition=None, deletion=None, admin=Fa
 
     message = "We analyzed %s regressions and found %s jobs to disable - " % (regression, jobs) + message
     if addition and deletion:
-        message += "\nThe jobs added to the current day compared to previous day are: ", addition
-        message += "\nThe jobs deleted from the current day compared to previous day are: ", deletion
+        message += "\nThe jobs added to the current day compared to previous day are:\n\n %s" % addition
+        message += "\n\nThe jobs deleted from the current day compared to previous day are:\n\n %s" % deletion
 
     msg = MIMEText(message)
     msg['Subject'] = "SETA Results - Date %s" % datetime.datetime.now().strftime("%Y-%m-%d")
--- a/tools/failures.py
+++ b/tools/failures.py
@@ -21,7 +21,7 @@ def getRawData(jobtype):
     return cdata['failures']
 
 def getDistinctTuples(jobtype):
-    platforms = ['osx10.6', 'winxp', 'osx10.8', 'win7','linux32','linux64']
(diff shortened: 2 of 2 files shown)


tools/failures.py
```diff
@@ -21,7 +21,7 @@ def getRawData(jobtype):
     return cdata['failures']
 
 def getDistinctTuples(jobtype):
-    platforms = ['osx10.6', 'winxp', 'osx10.8', 'win7','linux32','linux64']
+    platforms = ['osx10.6', 'osx-10-6', 'winxp', 'osx10.8', 'osx-10-8', 'win7', 'windows7-32', 'windowsxp', 'win8', 'windows8-64', 'linux32','linux64']
```
jmaher: this is a difference between tbpl and treeherder.  In looking into this I have modified the previous data  so we only have:
osx-10-6
osx-10-8
windowsxp
windows7-32
windows8-64
linux32
linux64

regarding android, we will need to do the same conversion, but I think for now we are good.  The buildtype and testtypes appear to be the same.

tools/emails.py
```diff
@@ -11,8 +11,8 @@ def send_email(regression, jobs, message, addition=None, deletion=None, admin=Fa
 
     message = "We analyzed %s regressions and found %s jobs to disable - " % (regression, jobs) + message
     if addition and deletion:
-        message += "\nThe jobs added to the current day compared to previous day are: ", addition
-        message += "\nThe jobs deleted from the current day compared to previous day are: ", deletion
+        message += "\nThe jobs added to the current day compared to previous day are:\n\n %s" % str(addition)
+        message += "\n\n\nThe jobs deleted from the current day compared to previous day are:\n\n %s" % str(deletion)
```
jmaher: why 3 \n ?  Also when using %s, I don't believe you need to str(var)

tools/emails.py
```diff
@@ -11,8 +11,8 @@ def send_email(regression, jobs, message, addition=None, deletion=None, admin=Fa
 
     message = "We analyzed %s regressions and found %s jobs to disable - " % (regression, jobs) + message
     if addition and deletion:
-        message += "\nThe jobs added to the current day compared to previous day are: ", addition
-        message += "\nThe jobs deleted from the current day compared to previous day are: ", deletion
+        message += "\nThe jobs added to the current day compared to previous day are:\n\n %s" % addition
+        message += "\n\nThe jobs deleted from the current day compared to previous day are:\n\n %s" % deletion
```
vaibhavmagarwal: Adding 2 '\n' to have a proper spacing in emails so that it becomes easier to read.

PR #43, opened 2015-01-26
create a system for defining a master list of tests

jmaher:
in order to provide accuracy, we need to have a master list of tests which we reference.  This is intended to be updated once daily, then queried as needed.


dminor:
lgtm

PR #44, opened 2015-01-28
update data retrieval, unique jobtype creation, and use unique jobtypes ...

jmaher:
...in failures.py/seta.py scripts.  This gets us almost perfect accuracy.

A few caveats:
1) if a new job shows up, we won't have a tbplcode for it
2) we need a script to auto update the jobtypes
3) the webUI isn't pretty yet
4) we now don't need all the meta info in the tbpltests json blob

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -384,7 +384,7 @@ def run_create_jobtypes_query():
         for buildtype in buildtypes:
             # ignore *build* types
             query = "select distinct testtype from testjobs where date>'%s'" % (twoweeks)
-            query += " and platform='%s' and buildtype='%s' and testtype not like '%%build%%'" % (platform, buildtype)
+            query += " and platform='%s' and buildtype='%s' and testtype not like '%%build%%' and testtype not like '%%_dep'" % (platform, buildtype)
             cursor = db.cursor()
             cursor.execute(query)
             types = []
@@ -397,7 +397,8 @@ def run_create_jobtypes_query():
                                 'tp5o', 'tp5o-e10s', 'dromaeojs', 'dromaeojs-e10s',
                                 'g1', 'g1-e10s', 'g1-snow', 'g1-snow-e10s', 
                                 'other_nol64', 'other_nol64-e10s', 'other_l64', 'other_l64-e10s',
-                                'xperf', 'xperf-e10s', 'dep', 'nightly', 'jetpack']:
+                                'xperf', 'xperf-e10s', 'dep', 'nightly', 'jetpack',
+                                'non-unified', 'valgrind']:
                     continue
 
                 if testtype:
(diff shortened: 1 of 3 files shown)


src/server.py
```diff
@@ -384,7 +384,7 @@ def run_create_jobtypes_query():
         for buildtype in buildtypes:
             # ignore *build* types
             query = "select distinct testtype from testjobs where date>'%s'" % (twoweeks)
-            query += " and platform='%s' and buildtype='%s' and testtype not like '%%build%%'" % (platform, buildtype)
+            query += " and platform='%s' and buildtype='%s' and testtype not like '%%build%%' and testtype not like '%%_dep'" % (platform, buildtype)
```
dminor: Please split this over multiple lines

src/server.py
```diff
@@ -458,12 +459,15 @@ def run_seta_details_query():
     date = request.args.get("date")
     db = create_db_connnection()
     cursor = db.cursor()
-    query = "select jobtype from seta where date='%s'" % date
+    query = "select jobtype from seta where date='%s 00:00:00'" % date
```
dminor: Is this left over from testing?

tools/failures.py
```diff
@@ -164,6 +165,7 @@ def run_query(query):
                          db="ouija")
 
     cur = db.cursor()
+    print query
```
dminor: Please remove (or write to the log file)

tools/seta.py
```diff
@@ -1,15 +1,16 @@
...
-                 'mochitest-other', 'reftest', 'xpcshell', 'crashtest', 'jsreftest', 'reftest-1', 'reftest-2',
-                 'reftest-e10s', 'crashtest-e10s', 'jittest', 'marionette', 'mochitest-gl', 'cppunit', 'reftest-no-accel',
-                 'web-platform-tests-1', 'web-platform-tests-2', 'web-platform-tests-3', 'web-platform-tests-4', 'web-platform-tests-reftests',
-                 'mochitest-e10s-browser-chrome-1', 'mochitest-e10s-browser-chrome-2', 'mochitest-e10s-browser-chrome-3',
-                 'mochitest-browser-chrome-1', 'mochitest-browser-chrome-2', 'mochitest-browser-chrome-3',
-                 'mochitest-devtools-chrome-1', 'mochitest-devtools-chrome-2', 'mochitest-devtools-chrome-3', 'mochitest-devtools-chrome']
-    buildtypes = ['debug', 'opt', 'asan']
-
-    return platforms, buildtypes, testtypes
+    conn = httplib.HTTPConnection('alertmanager.allizom.org')
```
dminor: We're using the requests library in updatedb.py to grab stuff, I think it would be better to use it here as well for consistency.

src/server.py
jmaher: this is actually the format it is in for the database.  I could do a followup and force this to just a yyyy-mm-dd and drop the 00:00:00 from the database and all related code.

dminor:
lgtm

PR #45, opened 2015-01-29
do some cleanup and future proofing for new jobs that are added and we don't know about them

jmaher:
This leaves 3 remaining tasks:
1) ensure we have a cronjob setup for create_jobtypes
2) update the webUI to be more informative
3) update the data in the emails to be more informative and point to the webUI report

### DIFF
--- a/tools/failures.py
+++ b/tools/failures.py
@@ -44,49 +44,49 @@ def format_in_table(active_jobs, master):
     sum_removed = 0
     sum_remaining = 0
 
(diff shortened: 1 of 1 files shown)


tools/failures.py
```diff
@@ -109,13 +110,22 @@ def format_in_table(active_jobs, master):
...
                 sum_removed += 1
             else:
                 output += "--"
                 sum_remaining += 1
 
         print "%s%s" % (key, output)
+
+    if missing_jobs:
+        print "** missing jobs: %s" % ','.join(missing_jobs)
+
```
dminor: This is a great idea, but instead of missing maybe we should call it unknown or unrecognized?

dminor:
Looks good!

PR #46, opened 2015-02-02
update failures to work on specific date, useful for larger iterations a...

jmaher:
...nd other investigations.  This isn't perfect, but with these changes I have the ability to do a lot of historical analysis.

### DIFF
--- a/tools/failures.py
+++ b/tools/failures.py
@@ -11,14 +11,18 @@ from emails import send_email
 
 import seta
 
-def getRawData():
-    url = "http://alertmanager.allizom.org/data/seta/"
+def getRawData(startDate, endDate):
+    if not startDate and not endDate:
+        url = "http://alertmanager.allizom.org/data/seta/"
+    else:
+        url = "http://alertmanager.allizom.org/data/seta/?startDate=%s&endDate=%s" % (startDate.strftime("%Y-%m-%d"), endDate.strftime("%Y-%m-%d"))
+
     response = requests.get(url, headers={'accept-encoding':'json'}, verify=True)
     data = json.loads(response.content)
     return data['failures']
 
 
-def communicate(failures, to_remove, total_detected, testmode, results=True):
+def communicate(failures, to_remove, total_detected, testmode, date, results=True):
 
     if results:
         active_jobs = seta.getDistinctTuples()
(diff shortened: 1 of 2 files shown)


tools/failures.py
```diff
@@ -11,14 +11,18 @@
 
 import seta
 
-def getRawData():
-    url = "http://alertmanager.allizom.org/data/seta/"
+def getRawData(startDate, endDate):
+    if not startDate and not endDate:
```
dminor: Not sure if this would actually be used, but if endDate is not specified it could default to today.

tools/failures.py
```diff
@@ -11,14 +11,18 @@
 
 import seta
 
-def getRawData():
-    url = "http://alertmanager.allizom.org/data/seta/"
+def getRawData(startDate, endDate):
```
dminor: I know we're not super consistent in Ouija, but these should probably be start_date and end_date.

tools/failures.py
```diff
@@ -113,7 +124,7 @@ def format_in_table(active_jobs, master):
                 try:
                     output += tbplnames[test]['code']
                 except KeyError:
-                    output += **
+                    output += '**'
```
dminor: Oops :)

tools/failures.py
```diff
@@ -130,31 +141,42 @@ def format_in_table(active_jobs, master):
...
 
-def sanity_check(failures, target):
+def sanity_check(failures, target, endDate=None):
     total = len(failures)
 
-    yesterday = datetime.datetime.now() - datetime.timedelta(days=1)
-    yesterday = yesterday.strftime('%Y-%m-%d')
+    if not endDate:
+        endDate = datetime.datetime.now()
+    endDate = endDate - datetime.timedelta(days=1)
```
dminor: Maybe make this midnight of the previous day?

tools/failures.py
```diff
@@ -244,22 +267,32 @@ def parse_args(argv=None):
...
-                to_remove, total_detected = sanity_check(failures, target)
-                if len(to_remove) > 0:
-                    communicate(failures, to_remove, total_detected, options.testmode, results=False)
-                    continue
-
-            # no quick option, or quick failed due to new failures detected
-            to_remove, total_detected = seta.depth_first(failures, target)
-            communicate(failures, to_remove, total_detected, options.testmode)
+        startDate = datetime.datetime(2014,8,13)
+        endDate = datetime.datetime(2014,11,13) # 3 months minimum
```
dminor: Maybe use options.start_date and options.end_date for this? I'm not sure what print_diff is doing above, but it doesn't seem that useful.

dminor:
lgtm

PR #47, opened 2015-02-04
remove --quick mode, fix some datetime vs str issues, make a single pass for much faster runs

jmaher:
now we run in <1 minute (locally) vs >20 minutes.


jmaher:
instead of running in 20 minutes we run in <1 minute - a typo in my original PR comment

PR #48, opened 2015-02-04
Adding detecting root cause functionality and ability to ignore jobs wit...

vaibhavmagarwal:
...h single failures.

### DIFF
--- a/tools/seta.py
+++ b/tools/seta.py
@@ -1,5 +1,6 @@
 import requests
 import json
+import copy
 
 def getDistinctTuples():
     url = "http://alertmanager.allizom.org/data/jobtypes/"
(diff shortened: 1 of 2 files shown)


tools/seta.py
```diff
@@ -85,7 +95,11 @@ def depth_first(failures, target):
     target = int(total* (target / 100))
     to_remove = []
 
-    to_remove = build_removals(active_jobs, failures, to_remove, target)
+    to_remove, remaining_failures = build_removals(active_jobs, failures, to_remove, target)
+    if ignoreFirstFailure:
+        remaining_jobtypes = [x for x in active_jobs if x not in to_remove]
```
jmaher: please add a comment here outlining the logic we are going to perform in this case.

jmaher:
Thanks Vaibhav!

PR #50, opened 2015-02-05
Cleaning up seta.py and making it pylint compatible.

vaibhavmagarwal:
Improving code quality such as removing camel case for uniformity and following standards. Also removing dead code like total_time and saved_time.


dminor:
Thanks for the cleanup!

PR #60, opened 2015-02-13
optional output for setadetails

jmaher:


src/server.py
```diff
@@ -468,11 +479,55 @@ def run_seta_details_query():
...
         parts = d[0].split("'")
         jobtype.append([parts[1], parts[3], parts[5]])
-    retVal[date] = jobtype
 
+    if active:
+        alljobs = jobtype_query()
+        active_jobs = []
+        for job in alljobs:
+            found = False
+            for j in jobtype:
```
dminor: isn't jobtype always an empty list at this point?

jmaher: jobtype is the array of jobtypes from the query (jobs we can run optionally), alljobs is the master list of all jobtypes, so here we are building a list of active_jobs which are alljobs-jobtype

dminor: Oh ok, I see it now

dminor:
Looks good to me

PR #63, opened 2015-03-10
remove id, get data from treeherder

jmaher:


src/updatedb.py
```diff
@@ -135,12 +135,16 @@ def uploadResults(data, branch, revision, date):
...
+        return
+
     job_property_names = data["job_property_names"]
     i = lambda x: job_property_names.index(x)
 
     results = data['results']
+    count = 0
     for r in results:
-        _id, log, slave, result, duration, platform, buildtype, testtype, bugid = '', '', '', '', '', '', '', '', ''
+        _id, logfile, slave, result, duration, platform, buildtype, testtype, bugid = '', '', '', '', '', '', '', '', ''
```
dminor: You can remove _id from here as well.

src/updatedb.py
```diff
@@ -181,29 +188,30 @@ def uploadResults(data, branch, revision, date):
...
+              (logfile, slave, _result, \
                duration, platform, buildtype, testtype, \
                bugid, branch, revision, date, failure_classification)
 
         try:
             cur.execute(sql)
+            count += 1
         except MySQLdb.IntegrityError:
             # we already have this job
+            logging.warning("sql failed to insert, we probably have this job: %s" % sql)
```
dminor: Without a primary key, this should never happen. I guess there is no harm in leaving this here in case we change things around later.

dminor:
Looks good to me.

PR #64, opened 2015-04-14
Adding functionality to display graphs and data for daily jobs/push to m...

vaibhavmagarwal:
...easure the impact of SETA

### DIFF
--- a/sql/dailyjobs.sql
+++ b/sql/dailyjobs.sql
@@ -1,0 +1,11 @@
+USE ouija;
+CREATE TABLE IF NOT EXISTS `dailyjobs` (
+    `date` datetime NOT NULL,
+    `platform` varchar(32) NOT NULL,
+    `branch` varchar(64) NOT NULL,
+    `numpushes` int NOT NULL,
+    `numjobs` int NOT NULL,
+    `sumduration` int NOT NULL,
+    index `date` (`date`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
+
--- a/src/server.py
+++ b/src/server.py
@@ -396,10 +396,10 @@ def run_create_jobtypes_query():
                 testtype = testtype[0]
                 # ignore talos, builds, jetpack
                 if testtype in ['svgr', 'svgr-e10s', 'svgr-snow', 'svgr-snow-e10s',
-                                'other', 'other-e10s', 'other-snow', 'other-snow-e10s', 
+                                'other', 'other-e10s', 'other-snow', 'other-snow-e10s',
                                 'chromez', 'chromez-e10s', 'chromez-snow', 'chromez-snow-e10s',
                                 'tp5o', 'tp5o-e10s', 'dromaeojs', 'dromaeojs-e10s',
-                                'g1', 'g1-e10s', 'g1-snow', 'g1-snow-e10s', 
+                                'g1', 'g1-e10s', 'g1-snow', 'g1-snow-e10s',
                                 'other_nol64', 'other_nol64-e10s', 'other_l64', 'other_l64-e10s',
                                 'xperf', 'xperf-e10s', 'dep', 'nightly', 'jetpack',
                                 'non-unified', 'valgrind']:
@@ -427,7 +427,9 @@ def run_seta_query():
 
     db = create_db_connnection()
     cursor = db.cursor()
-    query = "select bugid, platform, buildtype, testtype, duration from testjobs where failure_classification=2 and date>='%s' and date<='%s'" % (start_date, end_date)
+    query = "select bugid, platform, buildtype, testtype, duration from testjobs \
+             where failure_classification=2 and date>='%s' and date<='%s'" \
+             % (start_date, end_date)
     cursor.execute(query)
     failures = {}
     for d in cursor.fetchall():
@@ -527,6 +529,33 @@ def buildbot_name(platform, buildtype, jobname, branch):
 
     # TODO: do we need to do a jobname conversion?  I don't see a need yet
     return "%s %s %s test %s" % (platform_map[platform], branch, buildtype_map[buildtype], jobname)
+
+
+@app.route("/data/dailyjobs/")
+@json_response
+def run_dailyjob_query():
+    start_date, end_date = clean_date_params(request.args)
+    db = create_db_connnection()
+    cursor = db.cursor()
+    query = "select date, platform, branch, numpushes, numjobs, sumduration from dailyjobs where date>='%s' and date <='%s'" % (start_date, end_date)
+    cursor.execute(query)
+    output = {}
+    for rows in cursor.fetchall():
+        date = str(rows[0])
+        platform = rows[1]
+        branch = rows[2]
+        numpushes = int(rows[3])
+        numjobs = int(rows[4])
+        sumduration = int(rows[5])
+
+        if date not in output:
+            output[date] = {'mozilla-inbound': [], 'fx-team': []}
+        if 'mozilla-inbound' in branch:
+            output[date]['mozilla-inbound'].append([platform, numpushes, numjobs, sumduration])
+        elif 'fx-team' in branch:
+            output[date]['fx-team'].append([platform, numpushes, numjobs, sumduration])
+    return {'dailyjobs': output}
+
 
 @app.errorhandler(404)
 @json_response
--- a/static/dailyjobs.html
+++ b/static/dailyjobs.html
@@ -1,0 +1,161 @@
+<html>
+<head>
+  <script src="//ajax.googleapis.com/ajax/libs/jquery/2.0.3/jquery.min.js"></script>
+  <script src="//ajax.googleapis.com/ajax/libs/jqueryui/1.10.3/jquery-ui.min.js"></script>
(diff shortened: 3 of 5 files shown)


sql/dailyjobs.sql
```diff
@@ -0,0 +1,9 @@
+USE ouija;
+CREATE TABLE IF NOT EXISTS `dailyjobs` (
+    `date` datetime NOT NULL,
+    `platform` varchar(32) NOT NULL,
+    `branch` varchar(64) NOT NULL,
+    `numpushes` int NOT NULL,
+    `numjobs` int NOT NULL,
+    `sumduration` int NOT NULL
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```
dminor: I'm not certain how much data you're planning to retain in this table but you might want to consider creating indices on one or more of date, branch and platform to make the queries faster.

static/dailyjobs.html
```diff
@@ -0,0 +1,165 @@
...
+                <option id="windows"> Android </option>
+                </select>
+    </p>
+  <div id="placeholder" style="width:1200px;height:600px"></div>
+  <span id="hoverdata"></span>
+  <span id="clickdata"></span>
+  <script>
+    $(function() {
+      $( "#datepicker1" ).datepicker({
+        dateFormat: 'yy-mm-dd',
```
dminor: should this be 'yyyy-mm-dd' for consistency with the code below?

static/dailyjobs.html
```diff
@@ -0,0 +1,165 @@
...
+      start_date = date;
+      callJson(start_date, end_date);
+    }
+
+    function callEndDate(date){
+      end_date = date;
+      callJson(start_date, end_date);
+    }
+
+    function callJson(start_date, end_date){
```
dminor: This should be named something more specific than callJson, maybe getDailyJobsByDate?

static/dailyjobs.html
```diff
@@ -0,0 +1,165 @@
...
+      var url = "http://alertmanager.allizom.org/data/dailyjobs/?";
+      if (typeof start_date === 'undefined' || typeof end_date === 'undefined'){
+        url += "startDate=2015-01-01&endDate=2015-04-13";
```
dminor: I think endDate should default to today's date (or yesterday's if you want to ensure a full day's worth of data) and startDate should be a week less than endDate.

static/dailyjobs.html
```diff
@@ -0,0 +1,165 @@
...
+      } else {
+        url += "startDate="+start_date+"&endDate="+end_date;
+      }
+      $.getJSON(url, function(data){
+          dictionary = data["dailyjobs"];
```
dminor: This should be:
    var dictionary =

static/dailyjobs.html
```diff
@@ -0,0 +1,165 @@
...
+          var plot_data = [];
+          var ticks = [];
+
+          for (var key in data["dailyjobs"]){
+            console.log(platform);
```
dminor: I don't think you need this anymore :)

static/dailyjobs.html
```diff
@@ -0,0 +1,165 @@
...
+            "background-color": "#fee",
+            opacity: 0.80
+            }).appendTo("body");
+
+          $("#placeholder").bind("plothover", function (event, pos, item) {
+          if (item) {
+            var x = item.datapoint[0].toFixed(2),
+            y = item.datapoint[1].toFixed(2);
+            x = parseInt(x);
+            x = new Date(x);
```
dminor: I think this would be more readable if you used x = new Date(parseInt(x));

tools/dailyjobs_query.py
```diff
@@ -0,0 +1,126 @@
...
+                         db="ouija")
+cur = database.cursor()
+
+branches = ["mozilla-inbound", "fx-team"]
+platforms = ["linux", "osx", "win", "android"]
+
+
+def parse_args(argv=None):
+    parser = ArgumentParser()
+    parser.add_argument("-s", "--start_date",
```
dminor: Please use --start-date rather than --start_date, and --end-date rather than --end_date.

tools/dailyjobs_query.py
```diff
@@ -0,0 +1,126 @@
...
+                        dest="end_date",
+                        help="ending date."
+                        )
+
+    options = parser.parse_args(argv)
+    return options
+
+
+def updatedb(date, platform, branch, numpushes, numjobs, sumduration):
+    cur.execute('delete from dailyjobs where date="%s" and branch="%s" and platform="%s"' % (date, branch, platform))
```
dminor: Please break this into multiple lines.

tools/dailyjobs_query.py
```diff
@@ -0,0 +1,126 @@
...
+    query = 'insert into dailyjobs (date, platform, branch, numpushes, numjobs, sumduration) values ("{0}", "{1}", "{2}", {3}, {4}, {5})'
```
dminor: Please break this into multiple lines.

tools/dailyjobs_query.py
```diff
@@ -0,0 +1,126 @@
...
+
+            numpushes += 1
+            numjobs += value[1]
+            sumduration += value[2]
+        print (date, platform, branch, numpushes, numjobs, sumduration)
+        updatedb(date, platform, branch, numpushes, numjobs, sumduration)
+
+
+def retrievedb(branch, date):
+    query = "select revision,count(result),sum(duration),platform from testjobs where branch='{0}' and date like '{1}%' group by revision,platform;"
```
dminor: Please break this into multiple lines.

tools/dailyjobs_query.py
```diff
@@ -0,0 +1,126 @@
...
+        platform = rows[3]
+        found = False
+        for p in platforms:
+            if platform.startswith(p):
+                platform = p
+                found = True
+                break
+
+        # we can skip b2g/mulet or other platforms that don't match our core set
+        if not found:
```
dminor: You might want to consider issuing a warning somewhere if you see data for platforms you're skipping - this might avoid confusion later on.

tools/dailyjobs_query.py
```diff
@@ -0,0 +1,126 @@
...
+            numjobs += value[1]
+            sumduration += value[2]
+        print (date, platform, branch, numpushes, numjobs, sumduration)
+        updatedb(date, platform, branch, numpushes, numjobs, sumduration)
+
+
+def retrievedb(branch, date):
+    query = "select revision,count(result),sum(duration),platform from testjobs where branch='{0}' and date like '{1}%' group by revision,platform;"
+    query = query.format(branch, date)
+    print query
```
dminor: This looks like leftover debugging that can be removed.

tools/dailyjobs_query.py
```diff
@@ -0,0 +1,126 @@
...
+if __name__ == "__main__":
+    options = parse_args()
+
+    if type(options.start_date) == str:
+        options.start_date = datetime.datetime.strptime(options.start_date, "%Y-%m-%d").date()
+    if type(options.end_date) == str:
+        options.end_date = datetime.datetime.strptime(options.end_date, "%Y-%m-%d").date()
+
+    if options.start_date and not options.end_date:
+        options.end_date = options.start_date
```
dminor: Another thing you could check is whether the end_date is before the start_date and either issue an error or swap the dates.

dminor:
A few small things to fix, but overall, looks good!

static/dailyjobs.html
vaibhavmagarwal: the library only allows 'yy-mm-dd'.... putting 'yyyy-mm-dd' gives an error.

tools/dailyjobs_query.py
vaibhavmagarwal: Could you explain what should I do here? I did not get it...

static/dailyjobs.html
dminor: ok :)

src/server.py
```diff
@@ -528,6 +528,33 @@ def buildbot_name(platform, buildtype, jobname, branch):
...
     return "%s %s %s test %s" % (platform_map[platform], branch, buildtype_map[buildtype], jobname)
 
+
+@app.route("/data/dailyjobs/")
+@json_response
+def run_dailyjob_query():
+    start_date, end_date = clean_date_params(request.args)
+    db = create_db_connnection()
+    cursor = db.cursor()
+    query = "select date, platform, branch, numpushes, numjobs, sumduration from dailyjobs where date>='%s' and date <='%s'" % (start_date, end_date)
```
dminor: Please wrap this over two or more lines

tools/dailyjobs_query.py
dminor: You could keep track of each not found platform in a set, then write a warning to the console or log once you've gone through all the data that says something like: "unrecognized platform: ..." This is optional, it might not end up being that useful.

vaibhavmagarwal: I don't think printing that information is useful right now, and we are purposely skipping those jobs. I think by having comment here, we can always go back and print the skipped platforms if we want to.

dminor:
Looks good to me!

PR #65, opened 2015-04-16
Adding an annotation in the dailyjobs.html graph when SETA started

vaibhavmagarwal:


vaibhavmagarwal:
The annotation looks like this: http://imgur.com/x8gdLZH

static/dailyjobs.html
```diff
@@ -122,13 +119,36 @@
           var options = {
             grid: {
               hoverable: true,
-              clickable: true
+              clickable: true,
+              // Add a vertical line on the graph when SETA starts
+              markings: [{
+                color: "#000",
+                lineWidth: 1,
+                xaxis: { from: 1428600000000, to: 1428600000000 }
```
jmaher: ideally I would like a human date here like so:
xaxis: {from: ts('April 10, 2015'), to: ts('April 10, 2015')}

where ts takes the human date and converts to a timestamp.

alternatively, it would be nice to have a list if events:
var events = [{'date': '2015-04-10', 'title': 'Seta Starts'},{'date':'2015-06-31', 'title': 'Added Android'}]

then we could generate the markings and the placeholder and ctx.* calls in a loop.

PR #66, opened 2015-04-18
Adding weeks option to the dailyjobs.html graph.

vaibhavmagarwal:


static/dailyjobs.html
```diff
@@ -31,6 +35,20 @@
     </p>
   <div id="placeholder" style="width:1200px;height:600px"></div>
   <script>
+    Date.prototype.getWeekNumber = function(){
+      var d = new Date(+this);
+      d.setHours(0,0,0);
+      d.setDate(d.getDate()+4-(d.getDay()||7));
+      return Math.ceil((((d-new Date(d.getFullYear(),0,1))/8.64e7)+1)/7);
```
jmaher: can you comment on the logic behind the math here?

jmaher:
we could also send this into python to calculate at the server side.  Although I think this is fine.  Could this be extended for monthly as well?

vaibhavmagarwal:
Added the comment. Yes, this could be done on python side, but I thought it is better to take data from one api, rather than having multiple api's. And the same logic was going to be used on python side, so did that on javascript side, also could reuse a lot of existing code in javascript because of that. It could also be done monthly, I will do that in another PR.

jmaher:
looks good- we could use some polish on the X axis, but this is a nice view when looking at 6 weeks!

PR #76, opened 2015-09-10
Bug 1196934 - refactor SETA code to support all jobs types, including…

jmaher:
… new ones. r=vaibhav


tools/failures.py
```diff
@@ -194,7 +80,15 @@ def format_in_table(active_jobs, master):
...
+                    found = False
+                    for job in running_jobs:
+                        if job['name'] == test:
+                            output += job['job_type_symbol']
+                            found = True
+                            break
+
+                    if not found:
+                        raise Exception("unknown test: %s" % test)
                 except KeyError:
```
vaibhavmagarwal: The try..catch block can be removed since there will no longer be a KeyError as we are shifting to the 'found' variable approach.

static/js/seta.js
```diff
@@ -193,16 +170,18 @@ $(function() {
...
-                    'S': {'group': 'S'}
-                  }
+
+      var types = {'O': {}}; //Default with other
+      for (var jid = 0; jid < treenames.length; jid++) {
+          var g = treenames[jid]['job_group_symbol'];
+          if (g in types) {
+              continue;
+          }
+          types[g] = {};
```
vaibhavmagarwal: You can reduce this code to:
```
if ( !(g in types) ) {
    types[g] = {};
}
```

static/js/seta.js
```diff
@@ -215,17 +194,24 @@ $(function() {
...
+        for (var jid = 0; jid < treenames.length; jid++) {
+          if (treenames[jid]['name'] == active_osjobs[os][j]) {
+              group = treenames[jid]['job_group_symbol'];
+              jobcode = treenames[jid]['job_type_symbol'];
+              break;
+          }
+        }
+
+        if (jobcode != ''&& group == '') {
+           group = 'O';
```
vaibhavmagarwal: Missing space between '' and &&

src/server.py
```diff
@@ -533,6 +538,34 @@ def buildbot_name(platform, buildtype, jobname, branch):
...
     return "%s %s %s test %s" % (platform_map[platform], branch, buildtype_map[buildtype], jobname)
 
+@app.route("/data/jobnames/")
+@json_response
+def run_jobnames_query():
+    db = create_db_connnection()
+    cursor = db.cursor()
+    query = "select platform, buildtype, testtype from uniquejobs"
+    cursor.execute(query)
+    branch = 'mozilla-inbound'
```
vaibhavmagarwal: Any specific reason to hardcode branch as 'mozilla-inbound' here? Please add a comment if so.

src/server.py
```diff
@@ -533,6 +538,34 @@ def buildbot_name(platform, buildtype, jobname, branch):
     # TODO: do we need to do a jobname conversion?  I don't see a need yet
     return "%s %s %s test %s" % (platform_map[platform], branch, buildtype_map[buildtype], jobname)
 
+@app.route("/data/jobnames/")
+@json_response
+def run_jobnames_query():
+    db = create_db_connnection()
+    cursor = db.cursor()
+    query = "select platform, buildtype, testtype from uniquejobs"
```
vaibhavmagarwal: There is a function named jobtype_query(), you can remove a good part of repeated code here if you directly call that function.

src/server.py
```diff
@@ -533,6 +538,34 @@ def buildbot_name(platform, buildtype, jobname, branch):
...
+    cursor.execute(query)
+    branch = 'mozilla-inbound'
+    json_jobnames = {'results': []}
+    jobnames = json_jobnames['results']
```
vaibhavmagarwal: Delete this variable and directly use:
json_jobnames['results'].append(data) below.

src/server.py
```diff
@@ -395,14 +396,18 @@ def run_create_jobtypes_query():
...
+                                'g1', 'g2', 'g2-e10s', 'g1-e10s', 'g1-snow', 'g1-osx-e10s',
                                 'other_nol64', 'other_nol64-e10s', 'other_l64', 'other_l64-e10s',
-                                'xperf', 'xperf-e10s', 'dep', 'nightly', 'jetpack',
-                                'non-unified', 'valgrind', '5151c298eaed59034a45e3c4e3d4e0003fed4a14']:
+                                'xperf', 'xperf-e10s', 'other-e10s', 'dep', 'nightly', 'jetpack',
+                                'non-unified', 'valgrind']:
+                    continue
+
+                # skip taskcluster jobs
+                if len(testtype) == 40:
```
vaibhavmagarwal: I think this condition is not really required. What if the testtype for a non-taskcluster job is 40? It will then skip it even though we don't want that. Since there is no "b2g" in the platforms, I think this condition is unnecessary, let me know if I am missing something here.

jmaher: I am open to suggestions, but we have a lot of 40 character hash looking jobnames from task cluster.

vaibhavmagarwal:
@jmaher I would like if you could squash the two commits into a single commit (since you have repeated the content in both). Also, if you could mention the new jobtypes added that would be useful.

src/server.py
vaibhavmagarwal: Ok, we can leave this then.

tools/failures.py
```diff
@@ -53,123 +53,9 @@ def format_in_table(active_jobs, master):
...
-                 'robocop-3': {'group': 'RC', 'code': '3'},
-                 'robocop-4': {'group': 'RC', 'code': '4'},
-                 'robocop-5': {'group': 'RC', 'code': '5'},
-                 'robocop-6': {'group': 'RC', 'code': '6'},
-                 'robocop-7': {'group': 'RC', 'code': '7'},
-                 'robocop-8': {'group': 'RC', 'code': '8'},
-                 'reftest': {'group': 'R', 'code': 'R'},
-                 'androidx86-set-4': {'group': 'S', 'code': '4'}}
+    response = requests.get('http://alertmanager.allizom.org/data/jobnames/', headers={'accept-encoding': 'json'}, verify=True)
+    data = json.loads(response.content)
```
vaibhavmagarwal: requests has a .json() method, so you don't need to explicitly make json.loads
```
data = requests.get('http://alertmanager.allizom.org/data/jobnames/', headers={'accept-encoding': 'json'}, verify=True).json()
```
should work.

PR #77, opened 2015-09-14
Bug 1196935 - SETA code should have a list of jobs to consider, not a…

jmaher:
… list of jobs to ignore. r=vaibhav

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -372,8 +372,8 @@ def run_jobtypes_query():
 @app.route("/data/create_jobtypes/")
 @json_response
 def run_create_jobtypes_query():
-    # skipping b2g*, android*, mulet*
-    platforms = ['android-2-3-armv7-api9', 'android-4-2-x86', 'android-4-3-armv7-api11', 'osx-10-6', 'osx-10-10', 'windowsxp', 'windows7-32', 'linux32', 'linux64', 'windows8-64']
+    alljobs = jobtype_query()
+    platforms = list(set(job[0] for job in alljobs))
 
     # skipping pgo - We run this infrequent enough that we should have all pgo results tested
     buildtypes = ['debug', 'asan', 'opt']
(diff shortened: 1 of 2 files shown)


src/server.py
```diff
@@ -372,8 +372,12 @@ def run_jobtypes_query():
 @app.route("/data/create_jobtypes/")
 @json_response
 def run_create_jobtypes_query():
-    # skipping b2g*, android*, mulet*
-    platforms = ['android-2-3-armv7-api9', 'android-4-2-x86', 'android-4-3-armv7-api11', 'osx-10-6', 'osx-10-10', 'windowsxp', 'windows7-32', 'linux32', 'linux64', 'windows8-64']
+    alljobs = jobtype_query()
+    platforms = []
+    for job in alljobs:
+        if job[0] not in platforms:
+            platforms.append(job[0])
```
vaibhavmagarwal: This can be changed to a simple list comprehension:
```
platforms = list(set([job[0] for job in alljobs]))
```

src/server.py
```diff
@@ -372,8 +372,12 @@ def run_jobtypes_query():
...
+    print platforms
```
vaibhavmagarwal: Please print this with adequate information. eg:
```
print "The included platforms are: ", platforms
```

src/server.py
```diff
@@ -395,15 +399,8 @@ def run_create_jobtypes_query():
...
-                                'other', 'other-e10s', 'other-osx', 'other-osx-e10s',
-                                'chromez', 'chromez-e10s', 'chromez-osx', 'chromez-osx-e10s',
-                                'tp5o', 'tp5o-e10s', 'dromaeojs', 'dromaeojs-e10s',
-                                'g1', 'g2', 'g2-e10s', 'g1-e10s', 'g1-snow', 'g1-osx-e10s',
-                                'other_nol64', 'other_nol64-e10s', 'other_l64', 'other_l64-e10s',
-                                'xperf', 'xperf-e10s', 'other-e10s', 'dep', 'nightly', 'jetpack',
-                                'non-unified', 'valgrind']:
+                # ignore builds, jetpack
+                if testtype in ['dep', 'nightly', 'jetpack', 'non-unified', 'valgrind']:
                     continue
```
vaibhavmagarwal: I see you have removed talos jobs from consideration, do we want SETA active for talos jobs?

static/js/seta.js
```diff
@@ -140,16 +140,13 @@ $(function() {
...
-                         'osx-10-10 opt', 'osx-10-10 debug',
-                         'windowsxp opt', 'windowsxp debug',
-                         'windows7-32 opt', 'windows7-32 debug',
-                         'windows8-64 opt', 'windows8-64 debug',
-                         'android-2-3-armv7-api9 opt', 'android-4-2-x86 opt',
-                         'android-4-3-armv7-api11 opt', 'android-4-3-armv7-api11 debug'];
+    var active_oslist = [];
+
+    for (var os in active_osjobs) {
+      if (active_oslist.indexOf(os) < 0) {
```
vaibhavmagarwal: active_osjobs is a dictionary that is returned from buildOSJobMap. So we can be sure that each key is unique in it, therefore we do not need this if condition. Just directly do:
```
active_oslist.push(os);
```

PR #78, opened 2015-09-16
Bug 1196938 - SETA needs a way to have an over-ride list of jobs that…

jmaher:
… we can manually add/remove. r=vaibhav


src/preseed.json
```diff
@@ -0,0 +1,87 @@
...
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "chromez", "expires": "", "action": "run"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "chromez-e10s", "expires": "", "action": "coalesce"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "other", "expires": "", "action": "run"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "other-e10s", "expires": "", "action": "coalesce"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "g1", "expires": "", "action": "run"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "g1-e10s", "expires": "", "action": "coalesce"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "g2", "expires": "", "action": "run"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "g2-e10s", "expires": "", "action": "coalesce"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "dromaeojs", "expires": "", "action": "run"},
+ {"platform": "osx-10-10", "buildtype": "opt", "name": "dromaeojs-e10s", "expires": "", "action": "coalesce"}
```
vaibhavmagarwal: Why do you need the "expires" keyword? It is blank and I don't see it used anywhere. May as well remove it?

jmaher: the plan is for new tests which we need more data, that we will expire in
30 days.  Right now talos will be more static since we will not make
decisions based on historical data.

src/server.py
```diff
@@ -422,6 +416,25 @@ def run_create_jobtypes_query():
...
+               j[1] == job['buildtype'] and \
+               j[2] == job['name']:
+                found = True
+                break
+        if found:
+            continue
+
+        query = "insert into uniquejobs (platform, buildtype, testtype) values ('%s', '%s', '%s')" % (job['platform'], job['buildtype'], job['name'])
+        cursor.execute(query)
+
```
vaibhavmagarwal: This seems a bit wrong. Right now you have only talos jobs in preseed.json so you insert desktop jobs into uniquejobs before and then insert the talos jobs again here.

Instead insert talos jobs into jobtypes list directly and then insert into uniquejobs table together, that is a better non-confusing way.

src/server.py
```diff
@@ -11,7 +11,8 @@
 import MySQLdb
 from flask import Flask, request, json, Response, abort
 
-static_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "static"))
+SCRIPT_DIR = os.path.abspath(os.path.realpath(os.path.dirname(__file__)))
+static_path = os.path.abspath(os.path.join(SCRIPT_DIR, "..", "static"))
```
vaibhavmagarwal: This can be just:
```
SCRIPT_DIR = os.path.abspath(os.path.dirname(__file__))
static_path = os.path.join(os.path.dirname(SCRIPT_DIR), "static")
```

tools/failures.py
```diff
@@ -7,6 +8,7 @@
 
 import seta
 
+SCRIPT_DIR = os.path.abspath(os.path.realpath(os.path.dirname(__file__)))
```
vaibhavmagarwal: Here too:
```
SCRIPT_DIR = os.path.abspath(os.path.dirname(__file__))
```

tools/failures.py
```diff
@@ -231,6 +233,29 @@ def analyze_failures(start_date, end_date, testmode, ignore_failure, method):
         to_remove, total_detected = seta.failures_by_jobtype(failures, target, ignore_failure)
     else:
         to_remove, total_detected = seta.weighted_by_jobtype(failures, target, ignore_failure)
+
+    preseed_path = os.path.join(SCRIPT_DIR, '..', 'src', 'preseed.json')
```
vaibhavmagarwal: Use:
```
preseed_path = os.path.join(os.path.dirname(SCRIPT_DIR), 'src', 'preseed.json')
```

src/server.py
```diff
@@ -535,8 +550,13 @@ def buildbot_name(platform, buildtype, jobname, branch):
     if not branch:
         branch = "%s"
 
+    for talosname in ['tp5o', 'chromez', 'dromaeojs', 'other', 'g1', 'g2', 'xperf', 'svgr']:
+        if jobname.startswith(talosname):
+            buildtype = 'talos'
+            break
```
vaibhavmagarwal: startswith accepts a tuple of names to compare:
```
If jobname.startswith(('tp5o', 'chromez', 'dromaeojs', 'other', 'g1', 'g2', 'xperf', 'svgr')):
    buildtype = 'talos'
```

src/server.py
```diff
@@ -372,8 +373,8 @@ def run_jobtypes_query():
 @app.route("/data/create_jobtypes/")
 @json_response
 def run_create_jobtypes_query():
-    # skipping b2g*, android*, mulet*
-    platforms = ['android-2-3-armv7-api9', 'android-4-2-x86', 'android-4-3-armv7-api11', 'osx-10-6', 'osx-10-10', 'windowsxp', 'windows7-32', 'linux32', 'linux64', 'windows8-64']
+    alljobs = jobtype_query()
+    platforms = list(set(job[0] for job in alljobs))
```
vaibhavmagarwal: This was already changed and merged in https://github.com/dminor/ouija/pull/77
@jmaher How come it is not present in the repository??

src/server.py
jmaher: The problem here is that we have some talos jobs in the list already.  We
need to add all jobs from preseed and then all the jobs discovered from the
database.  Given that, we can do this in a different order, but we will
still have to check for duplicates.  Would you prefer me to add preseed
first and then check for duplicates while adding from the db, or keep it
the same?

vaibhavmagarwal: Checking for duplicates is fine. We insert in uniquejobs just before the additional code too: https://github.com/dminor/ouija/blob/ignorelist/src/server.py#L415
I would like to add all the jobs to jobtypes from preseed, and then insert in uniquejobs only once.

src/server.py
jmaher: hmm, it appears I messed up the landing on the master branch and never
actually merged it.  I am not sure how the previous pull request was merged
in:
https://github.com/dminor/ouija/commit/9efc6e7eb80b04a8a56ce8398e9ac1cae4422738

but not showing up.

src/server.py
```diff
@@ -417,6 +411,22 @@ def run_create_jobtypes_query():
...
+        found = False
+        for j in jobtypes:
+            if j[0] == job['platform'] and \
+               j[1] == job['buildtype'] and \
+               j[2] == job['name']:
+                found = True
+                break
+        if not found:
+            jobtypes.append([job['platform'], job['buildtype'], job['name']])
+
```
vaibhavmagarwal: You can just do:
```
if job not in jobtypes:
    jobtypes.append([job['platform'], job['buildtype'], job['name']])
```

tools/failures.py
```diff
@@ -231,6 +233,29 @@ def analyze_failures(start_date, end_date, testmode, ignore_failure, method):
...
+    for job in preseed:
+        #TODO: if expired, ignore
+
+        found = False
+        for item in to_remove:
+            if item[0] == job['platform'] and \
+               item[1] == job['buildtype'] and \
+               item[2] == job['name']:
+                found = True
+                break
```
vaibhavmagarwal: No need for the inner 'for' loop. Below you can do:
```
if job in to_remove and job['action'] == 'run':
    to_remove.remove(job)
elif job not in to_remove and job['action'] == 'coalesce':
    to_remove.append([job['platform'], job['buildtype'], job['name']])
```

src/server.py
```diff
@@ -535,8 +547,11 @@ def buildbot_name(platform, buildtype, jobname, branch):
     if not branch:
         branch = "%s"
 
+    if jobname.startswith(('tp5o', 'chromez', 'dromaeojs', 'other', 'g1', 'g2', 'xperf', 'svgr')):
+        buildtype = 'talos'
+
```
vaibhavmagarwal: @jmaher On thinking about this, why is buildtype set here for talos? buildtype is already passed to this function, is there a reason to believe that the buildtype is not correctly set?

PR #80, opened 2016-02-17
Bug 1248944 - Add a User Agent to requests made to Treeherder

edmorley:
In bug 1230222 we're encouraging consumers of Treeherder's API to set user agent:
https://bugzilla.mozilla.org/show_bug.cgi?id=1230222

As part of the refactoring to do this, I've removed `verify=True` since it's set to True by default:
http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification

The `accept-encoding: gzip` header is also unnecessary, since requests adds it by default:
https://github.com/kennethreitz/requests/blob/v2.9.1/requests/utils.py#L588

I've removed the `SSLError` try-catch, since there should not be problems verifying Treeherder's certificate. If there are, we shouldn't ignore them.

I've also added raise_for_status(), which raises if the status code is not 200. This shouldn't cause anything to fail that wouldn't have done already (since the json decode would have previously raised anyway), however at least now the root cause will be clearer. I'm happy to add try-except handling of that if you'd rather requests to fail more gracefully than they did previously.

It's also unclear why the html version of pushlog is being scraped, since there is a json version available - so I've added a TODO so this can be investigated later.

This PR also updated requests to 2.9.1, since it contains several security fixes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ MySQL-python==1.2.5
 Werkzeug==0.9.6
 argparse==1.2.1
 itsdangerous==0.24
-requests==2.3.0
+requests==2.9.1
 wsgiref==0.1.2
--- a/src/updatedb.py
+++ b/src/updatedb.py
@@ -10,6 +10,11 @@ from Queue import Queue
 
 import requests
 
+DEFAULT_REQUEST_HEADERS = {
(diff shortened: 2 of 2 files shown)


dminor:
Thanks!

edmorley:
Thank you for merging :-)

PR #82, opened 2016-05-06
add failure column

MikeLing:
Things become better now https://pastebin.mozilla.org/8869799 :)


jmaher:
thanks for this, I had probably done this manually.

PR #84, opened 2016-05-18
use fetch_json instead of pushlog endpoint

MikeLing:


MikeLing:
@jmaher r? :)

jmaher:
thank you, this is much cleaner- keep finding issues like this!

PR #85, opened 2016-05-24
Add two more platform into pmap

MikeLing:
@jmaher r?

### DIFF
--- a/src/updatedb.py
+++ b/src/updatedb.py
@@ -150,11 +150,12 @@ def uploadResults(data, branch, revision, date):
                    
         testtype = r[i("ref_data_name")].split()[-1]
         if r[i("build_system_type")] == "taskcluster":
-            #TODO: this is fragile, current platforms as of Jan 26, 2016 we see in taskcluster
             pmap = {"linux64": "Linux64",
                     "linux32": "Linux32",
                     "osx-10-7": "MacOSX64",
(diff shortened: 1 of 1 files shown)


src/updatedb.py
```diff
@@ -150,11 +150,12 @@ def uploadResults(data, branch, revision, date):
                    
         testtype = r[i("ref_data_name")].split()[-1]
         if r[i("build_system_type")] == "taskcluster":
-            #TODO: this is fragile, current platforms as of Jan 26, 2016 we see in taskcluster
             pmap = {"linux64": "Linux64",
                     "linux32": "Linux32",
                     "osx-10-7": "MacOSX64",
                     "gecko-decision": "gecko-decision",
+                    "b2g-device-image": "B2G",
+                    "mulet-linux64": "Mulet",
```
jmaher: I would like to intentionally skip these platforms, so if we add them, we should consider doing that so we can ignore them properly.  All B2G and Mulet builds are either being shut down or are only running until a certain time on a calendar is hit and then they will be shutdown.  Can you outline what coverage we plan to get with this?

MikeLing: hmm, not I actually haven't thought about that. I just saw that TODO issue and want to fix that. Then i query each of the platform and add it into the map. That's it.

jmaher:
closing this out as we have no specific need to worry about those platforms.

PR #87, opened 2016-05-25
Fix the pep8 error by flask8

MikeLing:


MikeLing:
I still got some errors that I don't know how to fix it.
<code>1) src/server.py:324:62: F821 undefined name 'result'
                app.logger.debug('UNRECOGNIZED RESULT: %s' % result)
</code>
This error comes from this line [1], I haven't found the 'result' been init at anywhere. Maybe it's **res** rather than **result**?

[1] https://github.com/dminor/ouija/blob/master/src/server.py#L319
<code>2)src/updatedb.py:106:13: F841 local variable 'push' is assigned to but never used
            push = None
</code>
Do we really need this push[2]?

[2]https://github.com/dminor/ouija/blob/master/src/updatedb.py#L100-L100?

<code>3)src/updatedb.py:148:30: F841 local variable 'revision_list' is assigned to but never used
</code>
Do we need this 'revision_list'[3]? I haven't found it been used in anywhere.

[3]https://github.com/dminor/ouija/blob/master/tools/seta.py#L96

<code>4)tools/seta.py:125:67: F821 undefined name 'master_root_cause'
        copy_failures = remove_root_cause_failures(copy_failures, master_root_cause)
</code>
Hmm, I have no ideal about this master_root_cause[4]. I found it's actually **None** and that function may error out when code comes to there. ( Because None is not iterable 
[4] https://github.com/dminor/ouija/blob/master/tools/seta.py#L122

f? @jmaher

jmaher:
1) yes, we should be logging 'res'
2) we can remove 'push'
3) we can remove 'revision_list'
4) master_root_cause- that is used in some cases where we are looking for different frequencies of failures.  Typically we try to find all root causes, but there are modes to ignore one instance of a failure or X instances of a failure, in that loop we will use master_root_causes.  In this specific spot, we do not use the returned variable master_root_cause, if we need to assign it to _ or some other throwaway variable that is fine.

src/server.py
```diff
@@ -130,18 +131,23 @@ def binify(bins, data):
...
+    """ This function returns the total failures/total jobs data per day for all platforms.
+        It is sending the data in the format required by flot.Flot is a jQuery package used
+        for 'attractive' plotting
+    """
 
     start_date, end_date = clean_date_params(request.args)
 
-    platforms = ['android4.0', 'android2.3', 'linux32', 'winxp', 'win7', 'win8', 'osx10.6', 'osx10.7', 'osx10.8']
+    platforms = ['android4.0', 'android2.3', 'linux32', 'winxp', 'win7', 'win8', 'osx10.6',
+                 'osx10.7', 'osx10.8']
```
jmaher: can you just make this one platform per line?

jmaher:
one small request, the rest looks great!

MikeLing:
Add tox support and a setup.py this time. Please tell me if it's not good to go, or need me rebase it ;)

Thank you Joel @jmaher

tox.ini
```diff
@@ -0,0 +1,31 @@
...
+    pytest-capturelog
+    Flask==0.10.1
+    Jinja2==2.7.3
+    MarkupSafe==0.23
+    MySQL-python==1.2.5
+    Werkzeug==0.9.6
+    argparse==1.2.1
+    itsdangerous==0.24
+    requests==2.9.1
+    wsgiref==0.1.2
```
jmaher: do we need all of these dependencies to make this run?  it seems like a lot.  I don't understand mysql-python, flask, jinja2, requests, wsgiref.  If that is required for running tox, lets add a comment if we can.

Also can we add a comment below to the [flake8] section where ignore is- I would like to know what those magic codes mean (such as D100, E731)

MikeLing: I just remove those lines :)

As to the flake8 session, them are error type we want o ignore(they are necessary I think).
e.g D103 require docstring in each public function,  D100 need docstring in each public module.
**One thing need to be mentioned is E731**. It's about we use lambda function in https://github.com/dminor/ouija/blob/master/src/updatedb.py#L130, which is E731 don't allow.
(you will see **src/updatedb.py:141:5: E731 do not assign a lambda expression, use a def if you remove it**) Do you want me fix that? I just thought we could forget it for now :)

@jmaher Thank you for you feedback! please review it again ;)

MikeLing:
@jmaher r?

setup.py
```diff
@@ -0,0 +1,23 @@
...
+required = [
+    "Flask",
+    "Jinja2",
+    "MarkupSafe",
+    "MySQL-python",
+    "Werkzeug",
+    "argparse",
+    "itsdangerous",
+    "requests",
+    "wsgiref",
```
jmaher: I would reduce the packages here as well!

jmaher:
regarding the specific error messages, lets try to change code to meet some of these requirements so we have a smaller list of excluded errors.

setup.py
MikeLing: hmm, I just copy and paste the whole 'requirements.txt'. I just want to ensure all the packages has been installed correctly before tox run. So, could you give me some advice that which package is what we need or which one isn't?

MikeLing:
@jmaher regard the error codes we still have:

**D100, D101,D102 and D103: Missing docstring in public module, class, method and function.**  I think we should give as much docstrings as we can, but I just don't need to do it for all of them(at least not now).

**D205: 1 blank line required between summary line and description.**  I didn't get that point, but I google it and it looks like most of people just ignore it.

**D400: First line should end with a period**. Something like "Parse request date params" will become Parse request date params**.**  I don't think it's a good idea

**D401: First line should be in imperative mood**. imperative mood? Seriously? :P

MikeLing:
@jmaher r?

jmaher:
thank you!

PR #88, opened 2016-06-09
Store high value job and store 90 days data

MikeLing:
Haven't ready yet, just for keep us in the same page :) @jmaher

### DIFF
--- a/static/js/seta.js
+++ b/static/js/seta.js
@@ -1,7 +1,7 @@
 $(function() {
   var input_date = location.search.substr(1).split('=')[1];
 
-  $error = $("#error"),
+  $error = $("#error");
   $body = $("body");
 
 
@@ -12,7 +12,7 @@ $(function() {
   }
 
   function createDates(data) {
-    var dates = {}
+    var dates = {};
 
     //NOTE: dates are in: "2015-01-01 00:00:00" format
     var previous_value = 0;
@@ -88,10 +88,10 @@ $(function() {
   function toggleState() {
     item = document.getElementById("toggle");
     fetchData();
-    if (item.value == "Show Optional Jobs") {
-      item.value = "Hide Optional Jobs";
+    if (item.value == "Show all Jobs") {
+      item.value = "Show required Jobs";
     } else {
-      item.value = "Show Optional Jobs";
+      item.value = "Show all Jobs";
     }
   }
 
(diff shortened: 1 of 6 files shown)


static/js/seta.js
```diff
@@ -146,15 +146,15 @@ $(function() {
     }
 
     var mytable = $('#seta');
-    var desc = "This is the list of jobs that would be required to run in order to catch every regression in the last 6 months";
+    var desc = "This is the list of jobs that would be required to run in order to catch every regression in the last 9 days";
```
jmaher: make this comment say 90 days instead of 9 :)

tools/failures.py
```diff
@@ -113,19 +113,28 @@ def format_in_table(active_jobs, master):
...
-    for jobtype in to_remove:
+    for jobtype in to_insert:
         query = 'insert into seta (date, jobtype) values ("%s", ' % date
         query += '"%s")' % jobtype
         run_query(query)
 
 
+def prepare_the_database():
+    # wipe up the job data older than 90 days
+    date = datetime.datetime.now() - datetime.timedelta(days=90)
```
jmaher: I would like us to make 90 a constant at the top of the file.

tools/failures.py
```diff
@@ -113,19 +113,28 @@ def format_in_table(active_jobs, master):
...
+    outdate_jobs = run_query("select jobtype from seta where date>'%s'" % date)
+    if len(outdate_jobs) > 0:
+        print "we have %s outdate jobs need to remove" % len(outdate_jobs)
+    run_query("delete from seta where date>='%s'" % date.strftime('%Y-%m-%d'))
```
jmaher: please verify the query is correct, I think we want less than, not greater than:
"delete from seta where date<='%s'"

tools/failures.py
MikeLing: Hmm, what should we named this constant? Just DAYS? (sorry, I'm really not good at naming

jmaher: I would name it: SETA_WINDOW=90

where window refers to a window in time :)

MikeLing:
@jmaher I still haven't add test for it, but I think it may good to go now(at least it looks same with the production version) So I ask r? in here  :)

jmaher:
looks great!

PR #89, opened 2016-06-13
get rid of logfile in ouija

MikeLing:


MikeLing:
@jmaher r?

sql/seta_schema.sql
```diff
@@ -2,7 +2,7 @@ USE ouija;
 
 CREATE TABLE IF NOT EXISTS `seta` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
-  `jobtype` varchar(64) NOT NULL,
+  `jobtype` varchar(1024) NOT NULL,
```
jmaher: why is this 1024 instead of 64?  can you give an example of a jobtype which is larger?

jmaher:
I would not want to check this into dminor's branch as the live version of
seta runs on ouija which runs other tools.

I am fine getting rid of the logfile for the main ouija toolchain.  The
main toolchain that uses ouija is the android failure rate- maybe we upload
all android and only failure_classification=2 for osx/windows/linux ?

sql/seta_schema.sql
MikeLing: It shows up one or two times when I insert data into seta table, I can't remember the special jobtype name now. But I could find it out later :)

MikeLing:
Alright, so do we need to get rid of the log file for the main ouija toolchain? I could update this PR and ask review again.

jmaher:
lets get rid of just the logfile, that will be a win in the database size.
We can work on a plan to only store the new stuff- maybe on heroku only?

MikeLing:
yeah, sounds great to me :)  Maybe we could start to work on teaching seta about taskcluster after that?

BTW, we could add stage server for seta by https://devcenter.heroku.com/articles/multiple-environments.

jmaher:
yes, the next step is taskcluster, I think the runnable api will be
helpful, but not required.  We just need to ensure that we are storing the
taskcluster jobs (linux64 debug *) properly in the database so we can match
up similar jobs.  Then the next part is making seta have an interface that
works with taskcluster- a few things to figure out.

MikeLing:
@jmaher r?

PR #90, opened 2016-06-22
fix the broken display

MikeLing:
The logic of display jobs should be:

[show all jobs]
  **display all required jobs**
[show required jobs]
   **display all job**


MikeLing:
@jmaher r?

PR #91, opened 2016-06-22
make jobs on linux64 debug displayable

MikeLing:


src/treecodes.py
```diff
@@ -213,15 +221,15 @@ def getGroup(name):
 
 def getGroupCode(name):
     try:
-        code = tbplnames[name]['group']
+        code = tbplnames[name.strip()]['group']
```
MikeLing: some tests' name have useless blank (e.g ' mochitest-devtools-chrome-e10s-4') and it make we can't get the group symbol.

MikeLing: please tell me if need a docstring for this :)

MikeLing:
Maybe we could do this better after runnable api been landed.

src/treecodes.py
```diff
@@ -42,11 +43,18 @@
...
              'mochitest-e10s-devtools-chrome-2': {'group': 'M-e10s', 'code': 'dt2'},
              'mochitest-e10s-devtools-chrome-3': {'group': 'M-e10s', 'code': 'dt3'},
              'mochitest-e10s-devtools-chrome-4': {'group': 'M-e10s', 'code': 'dt4'},
+             'mochitest-devtools-chrome-e10s-1': {'group': 'M-e10s', 'code': 'dt4'},
+             'mochitest-devtools-chrome-e10s-2': {'group': 'M-e10s', 'code': 'dt4'},
+             'mochitest-devtools-chrome-e10s-3': {'group': 'M-e10s', 'code': 'dt4'},
+             'mochitest-devtools-chrome-e10s-4': {'group': 'M-e10s', 'code': 'dt4'},
              'mochitest-push': {'group': 'M', 'code': 'p'},
              'xpcshell': {'group': '', 'code': 'X'},
              'xpcshell-1': {'group': '', 'code': 'X1'},
```
armenzg: As I mentioned few times, I believe this data structure can be generated automatically.
It is pretty standard string manipulation.

src/treecodes.py
armenzg: Please add a docstring to the function.

PR #94, opened 2016-07-06
send note only when new tests on the list

MikeLing:
Here is what I found,

If we have a previous job list(start_tuple) like [job1, job2, job3] and the job list we have now(end_tuple) is [job1, job2, job3, job4].

If we use start_tuple - end_tuple, the list we got will be [] and it's the reason why we only get notification when old job been removing rather than new one been added.

So, we could just inverse them which use end_tuple to minus the start_tuple to make it works when new one been added.

### DIFF
--- a/tools/failures.py
+++ b/tools/failures.py
@@ -189,10 +189,13 @@ def print_diff(start_date, end_date):
     else:
         deletion = list(set(start_tuple) - set(end_tuple))
         deletion.sort()
-        if not deletion:
+        if deletion:
+            print "%s: These jobs have changed from the previous day: %s" % \
+                  (end_date.strftime("%Y-%m-%d"), deletion)
+        else:
             deletion = ''
-        print "%s: These jobs have changed from the previous day: %s" % \
-              (end_date.strftime("%Y-%m-%d"), deletion)
+            print "%s: NO job been changed from the previous day" % \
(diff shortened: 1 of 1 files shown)


MikeLing:
@jmaher r?

jmaher:
I like this, can you test it on different date ranges to see what happens?  Ideally you could run this on a shorter timewindow with the existing heroku data (say 14 days), and use a recent SETA email to determine what changes we have and see if they show up.

MikeLing:
@jmaher Sorry for the late reply. I always curious about why the email is e.g :

The jobs **added** to the current day compared to previous day are:

 ["['linux32', 'debug', 'mochitest-devtools-chrome-4']"]

Why we call it 'added' rather than 'removed' if we only send note when old job been removed from list. And maybe we could figure out how to send different email to meet these two different situdation? like:

The jobs **added** to the current day compared to previous day when a new job been added and The jobs **removed** to the current day compared to previous day when a old job been removed.

jmaher:
yes, we should email both added/removed as separate lists.

armenzg:
Is there value on keeping this PR around? or can we close it?

MikeLing:
yep, I think we can close it for now due to I haven't any progress on it.

2016-10-19 22:39 GMT+08:00 Armen Zambrano <notifications@github.com>:

PR #99, opened 2016-07-09
fix the sql synatx

MikeLing:


sql/seta_schema.sql
```diff
@@ -2,7 +2,7 @@ USE ouija;
 
 CREATE TABLE IF NOT EXISTS `seta` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
-  `jobtype` varchar(64) NOT NULL,
+  `jobtype` varchar(1024) NOT NULL,
```
MikeLing: I meet **_mysql_exceptions.DataError: (1406, "Data too long for column 'jobtype' at row 1")** problem. Maybe change it to 512 or 256 is more reasonable, but 64 is too small anyway. :)

tools/failures.py
```diff
@@ -145,7 +145,7 @@ def prepare_the_database():
     if len(outdate_jobs) > 0:
         print "we have %s outdate jobs need to remove" % len(outdate_jobs)
         for job in outdate_jobs:
-            run_query("delete from seta where jobtype=%s" % job)
+            run_query("delete from seta where jobtype='%s' and date<='%s'" % (job, date))
```
MikeLing: I keep the outdate_jobs because I think people may want to know we are going to delete data for them. But it has a  little redundance. Maybe we could find another way to fix that?

sql/seta_schema.sql
jmaher: 256 should be good enough.

tools/failures.py
jmaher: this is a script which runs in the background, no need to make it verbose.

PR #101, opened 2016-07-12
separate taskcluster jobs from buildbot jobs

MikeLing:
Fix some flake8 error in this patch also ;)

### DIFF
--- a/src/jobtypes.py
+++ b/src/jobtypes.py
@@ -5,10 +5,9 @@ from tools.failures import update_runnableapi
 LOG = logging.getLogger(__name__)
 JSONPATH = os.getcwd() + '/runnablejobs.json'
 
+
 def _getgroup(name):
-    """
-    It's about to get group for the test job, like crashtest for crashtest-1.
-    """
+    """It's about to get group for the test job, like crashtest for crashtest-1."""
     try:
         group = name.split('-')[0]
     except:
@@ -68,11 +67,32 @@ class Treecodes:
                 #  opt test web-platform-tests-4"
                 if job['build_system_type'] == 'buildbot':
                     testtype = job['ref_data_name'].split(' ')[-1]
+                    job_identifier = job['ref_data_name']
 
                 # taskcluster's testtype is a part of its 'job_type_name' like reftest-2
                 # for [TC] Linux64 reftest-2
                 else:
-                    testtype = job['job_type_name'].split(' ')[-1]
+                    # The test name on taskcluster comes to a sort of combination
+                    # (e.g desktop-test-linux64/debug-jittests-3) and asan job can
+                    # been referenced as a opt job. BTW, job like
+                    # 'MacOSX64 Static Analysis Opt' will been leave aside because
+                    # they are not test job anyway.
+                    if job['ref_data_name'].startswith('desktop-test') or \
+                            job['ref_data_name'].startswith('android-test'):
+
+                        # we want the build type(debug or opt) to separate the job_type_name
+                        # (e.g desktop-test-linux64/debug-jittests-3)
+                        separator = job['platform_option'] \
+                            if job['platform_option'] != 'asan' else 'opt'
+                        # we should get "jittests-3" as testtype for job_type_name like
+                        # desktop-test-linux64/debug-jittests-3
+                        testtype = job['job_type_name'].split(
+                            '{buildtype}-'.format(buildtype=separator))[-1]
+                        job_identifier = job['ref_data_name']
+
+                    # we only care about the test job.
+                    else:
+                        continue
 
                 # we don't need non-test job for seta, but we still need to verify this type list
                 if testtype in ['dep', 'nightly', 'non-unified', 'valgrind', 'build']:
@@ -90,7 +110,7 @@ class Treecodes:
                         # It's about get jobnames and both buildbot and taskcluster job has '?'
                         # when the job_group_symbol is unknown.
                         self.jobnames.append(self._get_jobnames(job, testtype,
-                                                                job['ref_data_name']))
+                                                                job_identifier))
                         job_group_symbol = job['job_group_symbol'] \
                             if job['job_group_symbol'] != '?' else ''
                         job_type_symbol = job['job_type_symbol'] if job['job_type_symbol'] else ''
@@ -107,19 +127,20 @@ class Treecodes:
         """Query all jobnames including buildtype and groupcode, then return them as list"""
         return self.jobnames
 
-    def _get_jobnames(self, job, testtype, buildername):
+    def _get_jobnames(self, job, testtype, job_identifier):
         signature = ''  # TH specific
+        buildplatform = job['build_system_type']
         buildtype = job['platform_option']
         platform = job['platform']
         name = testtype
         jobname = ''  # TH specific
-        buildername = buildername
+        job_identifier = job_identifier
         group = _getgroup(name)
         groupcode = _getgroupCode(self.tbplnames, name)
         code = _getcode(self.tbplnames, name)
-        data = {'jobname': jobname, 'signature': signature,
+        data = {'buildplatform': buildplatform, 'jobname': jobname, 'signature': signature,
                 'job_type_name': group, 'job_group_symbol': groupcode,
                 'name': name, 'job_type_symbol': code,
-                'ref_data_name': buildername,
+                'ref_data_name': job_identifier,
                 'platform': platform, 'buildtype': buildtype}
         return data
--- a/src/server.py
+++ b/src/server.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 import os
 import calendar
 from datetime import datetime, timedelta
@@ -417,7 +415,8 @@ def run_seta_details_query():
     active = request.args.get("active", 0)
     buildbot = request.args.get("buildbot", 0)
     branch = request.args.get("branch", '')
-
+    taskcluster = request.args.get("taskcluster", 0)
+    jobnames = JOBSDATA.jobnames_query()
     if date == "" or date == "latest":
         today = datetime.now()
         date = today.strftime("%Y-%m-%d")
@@ -429,6 +428,10 @@ def run_seta_details_query():
     retVal = {}
     retVal[date] = []
     jobtype = []
+
+    # we only support fx-team and mozilla-inbound branch in seta
+    if (str(branch) in ['fx-team', 'mozilla-inbound', 'autoland']) is not True and str(branch) != '':
+        abort(404)
     for d in cursor.fetchall():
         parts = d[0].split("'")
         jobtype.append([parts[1], parts[3], parts[5]])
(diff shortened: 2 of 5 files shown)


MikeLing:
you can try it on heroku remote by: [buildbot jobs](http://seta-dev.herokuapp.com/data/setadetails/?buildbot=1) and [taskcluster](http://seta-dev.herokuapp.com/data/setadetails/?taskcluster=1). BTW, we only have failures for today. And I had try it on local environment already. :)

@jmaher r?

src/server.py
```diff
@@ -448,51 +447,28 @@ def run_seta_details_query():
...
+            for j in taskcluster_jobs:
+                if j['name'] == job[2] and j['platform'] == job[0] and j['buildtype'] == job[1]:
+                    active_jobs.append(j['ref_data_name'])
         jobtype = active_jobs
 
     retVal[date] = jobtype
     return {'jobtypes': retVal}
 
 
-# we still need this for builbot platform mapping
```
MikeLing: we don't need it anymore because it can get buildername from data/jobnames api directly.

src/server.py
```diff
@@ -448,51 +447,28 @@ def run_seta_details_query():
 
     if buildbot:
         active_jobs = []
+        # pick up buildbot jobs from job list to faster the filter process
+        buildbot_jobs = [job for job in jobnames if job['buildplatform'] == 'buildbot']
+        # find out the correspond job detail information
+        for job in jobtype:
+            for j in buildbot_jobs:
+                if j['name'] == job[2] and j['platform'] == job[0] and j['buildtype'] == job[1]:
+                    active_jobs.append(j['ref_data_name'])
```
MikeLing: Honestly, I don't know why we need to return buildername rather than jobtype[platform, build type, test type]. But just for keeping the same with  what seta returns here before, I make it return buildername at here and for taskcluster also.

src/jobtypes.py
```diff
@@ -90,7 +91,7 @@ def __init__(self, repo_name='mozilla-inbound'):
                         # It's about get jobnames and both buildbot and taskcluster job has '?'
                         # when the job_group_symbol is unknown.
                         self.jobnames.append(self._get_jobnames(job, testtype,
-                                                                job['ref_data_name']))
+                                                                buildername))
```
MikeLing: It's my mistake, but luckily we found it and make it right.

src/jobtypes.py
```diff
@@ -68,11 +67,13 @@ def __init__(self, repo_name='mozilla-inbound'):
                 #  opt test web-platform-tests-4"
                 if job['build_system_type'] == 'buildbot':
                     testtype = job['ref_data_name'].split(' ')[-1]
+                    buildername = job['ref_data_name']
 
                 # taskcluster's testtype is a part of its 'job_type_name' like reftest-2
                 # for [TC] Linux64 reftest-2
                 else:
                     testtype = job['job_type_name'].split(' ')[-1]
+                    buildername = job['job_type_name']
```
armenzg: IMHO it is better to name variables that are agnostic to both systems.
In Buildbot we have buildernames while on TaskCluster we have tasks labels.
A variable that would be indepdent to both systems would be job_identifier or job_uuid.

Note that we also have task label (which are unique) instead of task names (which is defined on the tree but the developer can by mistake name two tasks with the same name).

src/jobtypes.py
```diff
@@ -111,13 +112,14 @@ def _get_jobnames(self, job, testtype, buildername):
         signature = ''  # TH specific
         buildtype = job['platform_option']
         platform = job['platform']
+        buildplatform = job['build_system_type']
```
armenzg: Why not build_system_type? I fine it better to keep consistent naming of concepts across multiple code bases.

MikeLing: ok, let's keep it as 'build_system_type'. Thank you armenzg :)

src/jobtypes.py
MikeLing: hmm, what's the meaning of **agnostic to both systems** ? Is that mean we should create a new variable name which doesn't belong to Buildbot and TaskCluster either? How about **job_label**? I always feel lost with variable naming.

armenzg:
@jmaher @MikeLing I have some concerns with using task names to be used as part of SETA.

Unfortunately, engineers can name their tasks incorrectly yet not cause trouble anywhere on Treeherder. In other words, no one will notice that there is problem in the naming of tasks.
For instance, you could have a task for opt be named exactly the same as a debug task and SETA won't be able to know that they're different.

This problem goes away with task labels.
From push to push task labels refer to the same job. Always.

I spoke with dustin and the system by design guarantees this.
If we delete a task, the label will cease.
If we change parameters for a task the task label will still refer to the same task.

armenzg:
FYI dustin will be landing a bug to refactor the legacy tasks using the new system.
Only then we will have proper task labels that are stable from push to push.

armenzg:
I also want to point out that the number of tasks in here is very small to what it actually runs on production.
http://seta-dev.herokuapp.com/data/setadetails/?taskcluster=1

src/jobtypes.py
armenzg: You got the idea.

job_identifier is probably good to get the idea across.

MikeLing:
>I also want to point out that the number of tasks in here is very small to what it actually runs on >production. http://seta-dev.herokuapp.com/data/setadetails/?taskcluster=1

hmm, I don't think so. Because we aren't running failure.py everyday on heroku, due to the database size limit, we don't have so much failure data at first place. For example, if you visit http://seta-dev.herokuapp.com/data/setadetails/ (today)  we have **115** failures been detected. Then http://seta-dev.herokuapp.com/data/setadetails/?buildbot=1 can return **107** failures and http://seta-dev.herokuapp.com/data/setadetails/?taskcluster=1 can give you **8** failures.  So I think it's just go :)

armenzg:
I understand now. Thanks!

https://github.com/mozilla/ouija/pull/101#issuecomment-232089505 still applies. I don't want it to get lost.

src/server.py
jmaher: buildbot needs the specific buildernames, but that is only for the api endpoint.

jmaher:
I think #101 is important, SETA depends on 3 pieces of data:
platform (this needs to be uniform to some degree)
buildtype (opt/debug, could be pgo/asan/etc.)
testtype - this is the concern here.

SETA requires that we have consistent names (i.e. 'crashtest-4') for a given platform so that we can see if 'crashtest-4' has ever found a regression and when it did.  As you can imagine this is a hard problem to solve, if someone creates an invalid name, that is a bug.  If someone changes a name, we need to either update SETA or accept this as a brand new testtype.

src/server.py
```diff
@@ -416,7 +414,8 @@ def run_seta_details_query():
     date = request.args.get("date", "")
     active = request.args.get("active", 0)
     buildbot = request.args.get("buildbot", 0)
-    branch = request.args.get("branch", '')
```
jmaher: why is 'branch' removed?

src/server.py
```diff
@@ -448,51 +447,28 @@ def run_seta_details_query():
...
-    platform_map['android-4-3-armv7-api15'] = "android-4-3-armv7-api15"
-    platform_map['osx-10-10'] = "Rev7 MacOSX Yosemite 10.10.5"
-    platform_map['osx-10-8'] = "Rev5 MacOSX Mountain Lion 10.8"
-    platform_map['osx-10-6'] = "Rev4 MacOSX Snow Leopard 10.6"
-    platform_map['linux32'] = "Ubuntu VM 12.04"
-    platform_map['linux64'] = "Ubuntu VM 12.04 x64"
-    platform_map['linux64asan'] = "Ubuntu ASAN VM 12.04 x64"
-    platform_map['windowsxp'] = "Windows XP 32-bit"
-    platform_map['windows7-32'] = "Windows 7 32-bit"
-    platform_map['windows8-64'] = "Windows 8 64-bit"
```
jmaher: I think we might run into problems removing this platform_map.

src/server.py
```diff
@@ -448,51 +447,28 @@ def run_seta_details_query():
...
-        platform = 'linux64asan'
-
-    if not branch:
-        branch = "%s"
-
-    if jobname.startswith(('tp5o', 'chromez', 'dromaeojs', 'other', 'g1', 'g2', 'xperf', 'svgr')):
-        buildtype = 'talos'
-
-    # TODO: do we need to do a jobname conversion?  I don't see a need yet
-    return "%s %s %s %s" % (platform_map[platform], branch, buildtype_map[buildtype], jobname)
```
jmaher: why is all the buildtype stuff removed?

src/updatedb.py
```diff
@@ -84,7 +84,7 @@ def getCSetResults(branch, revision):
     num_results = 0
     retVal = {}
     while not done:
-        url = "https://treeherder.mozilla.org/api/project/%s/jobs/?count=%s&offset=%s&result_set_id=%s&return_type=list
+        url = "https://treeherder.mozilla.org/api/project/%s/jobs/?count=%s&offset=%s&result_set_id=%s&return_type=list"
```
jmaher: thanks for the fix here, please remove return_type=list from here.

src/server.py
MikeLing: @jmaher , sorry for the late reply. As I said in https://github.com/mozilla/ouija/pull/101/files#r70400656, the whole builder name been removed because I think we don't need it to return buildername for us right now.  Due to buildername has become a part of data returned by jobnames endpoint right now(https://github.com/mozilla/ouija/pull/101/files#diff-715c427ce6ef441fcf61777465fe6789R118). We could just pick up the correspond job from the jobnames list and return the right buildername. So I don't think it will make us run into trouble. Does it make sense to you?

src/server.py
MikeLing: hmm, because I think the branch in here is about to generate a buildername. e.g mozilla-inbound for Ubuntu VM 12.04 x64 **mozilla-inbound** opt test mochitest-devtools-chrome-3. But we could get it from jobnames endpoint directly. (Actually, we don't have too much choice for now, we only get mozilla-inbound data from runnable job endpoint.

src/server.py
MikeLing: hmm, maybe. But the platform map is somehow hardcode in here. So I think we just use activedata  to replace them.

jmaher: Earlier this week windows-7-vm was added to the list to fix an issue.  I
don't know the full story, but it seems we might be overlooking something
here.

src/server.py
jmaher: keep in mind that seta runs on fx-team and mozilla-inbound, probably soon to be on autoland.  We will have the same data for all branches.

src/server.py
MikeLing: yeah, it always hard to trace every platform and mapping them in here. So I can replace them by active data maybe is a better choice.

src/server.py
MikeLing: alright, that's really a issue need to solve. I will figure it out within these two day and add it into this patch.

MikeLing:
It can return a list of tasklabel in taskcluster case and a list of buildername for buildbot :) you can try it on heroku by today's data http://seta-dev.herokuapp.com/data/setadetails/?data=2016-07-20&buildbot=1 and http://seta-dev.herokuapp.com/data/setadetails/?data=2016-07-20&taskcluster=1

MikeLing:
Hey @jmaher and @armenzg , now it returns a list of 'tasklabel'. r? :)

screen shot 2016-07-22 at 2 42 39 pm

jmaher:
Nice !  I will review the code later, probably Sunday or Monday!

src/jobtypes.py
```diff
@@ -68,11 +67,27 @@ def __init__(self, repo_name='mozilla-inbound'):
...
                 if job['build_system_type'] == 'buildbot':
                     testtype = job['ref_data_name'].split(' ')[-1]
+                    job_identifier = job['ref_data_name']
 
                 # taskcluster's testtype is a part of its 'job_type_name' like reftest-2
                 # for [TC] Linux64 reftest-2
                 else:
-                    testtype = job['job_type_name'].split(' ')[-1]
+                    # we don't want job of build kind in seta
+                    if job['job_type_symbol'] == "I":
```
armenzg: This is not the only build job type.

src/jobtypes.py
```diff
@@ -68,11 +67,27 @@ def __init__(self, repo_name='mozilla-inbound'):
...
+                        continue
+                    else:
```
armenzg: If you want, you could whitelist test jobs instead of blacklisting build jobs.

For now, you can check that the label starts with 'desktop-test'.
In the future, we would have to add other platforms.

armenzg:
This lgtm.

src/jobtypes.py
MikeLing: Great! I will keep that in mind :)

jmaher: I like the idea of desktop-test as a filter to find test jobs, we might need android-test as well.

src/jobtypes.py
```diff
@@ -68,11 +67,27 @@ def __init__(self, repo_name='mozilla-inbound'):
...
+                        # and asan job can been referenced as a opt job. BTW, job like
+                        # 'MacOSX64 Static Analysis Opt' will been leave aside because
+                        # they are not test job anyway
+
+                        separator = job['platform_option'] \
+                            if job['platform_option'] != 'asan' else 'opt'
+
+                        testtype = job['job_type_name'].split(
+                            '{buildtype}-'.format(buildtype=separator))[-1]
+                        job_identifier = job['ref_data_name']
```
jmaher: can you give a couple examples of what data comes in and what we put in separator/testtype/job_identifier?

src/server.py
```diff
@@ -429,6 +428,10 @@ def run_seta_details_query():
     retVal = {}
     retVal[date] = []
     jobtype = []
+
+    # we only support fx-team and mozilla-inbound branch in seta
+    if (str(branch) in ['fx-team', 'mozilla-inbound']) is not True and str(branch) != '':
```
jmaher: please add 'autoland' as a third branch here!

src/server.py
jmaher: I am still concerned here that we won't have the same 'platform' between taskcluster/buildbot, possible this is something we just live with :)

MikeLing: hmmm, could you tell me about your concern more specifically? Because AFAIK we just query the platform from runnablejobs api directly.

src/server.py
MikeLing: my bad, sorry I forget it…

src/server.py
jmaher: to find all related jobs (similar, etc.) we need to have a matching platforms, buildtype, testtype values.  We had previously sanitized the platform names to be stuff like linux32, linux64, win32, win64.  Now we have buildbot values which are not guaranteed to be == to taskcluster values for platform.  It is ok to change the platform names in general, but we need some high degree of confidence that they won't change if somebody fiddles with a random config in-tree.

for the buildtype, this is important as well- right now we force things to opt/debug and treat asan as opt.  ideally we should be able to have opt/debug/asan- but then do we ignore other types like valgrind and pgo (I believe we should)?  If the names are not uniform between platforms or CI systems, then we need to force them to be similar by renaming them in SETA.

MikeLing: hmm, so you mean we should uniform the platform and buildtype to avoid people using the wrong config parameter? I'm not familiar with the configuration and way to trigger a push. AFAIK, the task schedule in taskcluster for push is automatically happened. However, I think if a man config a task/job with random(wrong or no-formal) parameter, the buildbot and taskcluster should return error in the  first place. :)

src/jobtypes.py
MikeLing: ok, I will work on that :)

src/server.py
jmaher: doing a search on ouija (which is using runnable_jobs.json) I don't see platform conflicts nor buildtype issues, I suspect this concern I have is primarily a worry, but nothing specific.  We do need to fix testtype though, we are currently storing tasklabels which are not useful for comparing between platforms.

MikeLing: > We do need to fix testtype though, we are currently storing tasklabels which are not useful for comparing between platforms.

hmmm, I don't think so.  If you try [jobnames](http://seta-dev.herokuapp.com/data/jobnames/) and [jobtypes](http://seta-dev.herokuapp.com/data/jobtypes/) endpoint on the heroku remote, you will see the 'name' parameter in /jobnames is the testtype we want and we actually have the appropriate testtype in /jobtypes endpoint. The only place we store tasklabel is the 'ref_data_name' .

MikeLing:
@jmaher r?

src/jobtypes.py
```diff
@@ -68,11 +67,31 @@ def __init__(self, repo_name='mozilla-inbound'):
...
+                    job_identifier = job['ref_data_name']
 
                 # taskcluster's testtype is a part of its 'job_type_name' like reftest-2
                 # for [TC] Linux64 reftest-2
                 else:
-                    testtype = job['job_type_name'].split(' ')[-1]
+                    # we don't want job of build kind in seta
+                    if job['ref_data_name'].startswith('desktop-test') or \
+                            job['ref_data_name'].startswith('android-test'):
+                        continue
```
jmaher: I think the logic here is backwards, we want all tests that start with desktop-test and android-test

PR #109, opened 2016-07-24
add sqlalchemy support

MikeLing:


MikeLing:
Hey @jmaher , now we not longer need to expose our sql command into the code. And the greatest thing is the data insert become pretty fast than before :)

please give me feedback :)

src/server.py
```diff
@@ -58,6 +53,7 @@ def wrapper(*args, **kwargs):
 
 
 def get_date_range(dates):
+    print dates
```
jmaher: this looks like debugging

src/server.py
```diff
@@ -148,17 +144,13 @@ def run_results_day_flot_query():
...
     for platform in platforms:
-        cursor = db.cursor()
-        cursor.execute("""select DATE(date) as day,sum(result="%s") as failures,count(*) as
-                          totals from testjobs where platform="%s" and date >= "%s" and date <= "%s"
-                          group by day""" % ('testfailed', platform, start_date, end_date))
-
-        query_results = cursor.fetchall()
-
+        query_results = session.query(Testjobs.date.label('day'),
+                                      func.count(Testjobs.result).label("failures"),
```
jmaher: did we change 'testfailed' -> 'failures'?  I don't know where we get the 'testfailed' value for the result field?

src/server.py
```diff
@@ -211,8 +194,9 @@ def run_slaves_query():
     summary = {result: 0 for result in labels}
     summary['jobs_since_last_success'] = 0
     dates = []
-
+    print query_results
     for name, result, date in query_results:
+        print name, result, date
```
jmaher: we are printing stuff here in line 197 and 199, that looks like debugging.

src/server.py
```diff
@@ -377,13 +352,12 @@ def run_jobtypes_query():
...
 
-    db = create_db_connnection()
-    cursor = db.cursor()
-    query = "select bugid, platform, buildtype, testtype, duration from testjobs \
-             where failure_classification=2 and date>='%s' and date<='%s'" % (start_date, end_date)
-    cursor.execute(query)
+    data = session.query(Testjobs.bugid, Testjobs.platform, Testjobs.buildtype, Testjobs.testtype,
+                         Testjobs.duration).filter(and_(Testjobs.failure_classification == 2,
+                                                        Testjobs.date >= start_date,
+                                                        Testjobs.date <= end_date)).all()
```
jmaher: possibly we could just use date between here, I don't know why we couldn't- possibly you could explain what the difference of including the dates vs just between would be?

src/server.py
```diff
@@ -507,19 +473,18 @@ def run_jobnames_query():
...
-                end" % (start_date, end_date)
-    cursor.execute(query)
+
+    data = session.query(Dailyjobs.date, Dailyjobs.platform, Dailyjobs.branch, Dailyjobs.numjobs,
+                         Dailyjobs.sumduration).\
+        filter(Dailyjobs.date.between(start_date, end_date)).order_by(case([
+            (Dailyjobs.platform == 'linux', 1),
+            (Dailyjobs.platform == 'osx', 2),
+            (Dailyjobs.platform == 'win', 3),
+            (Dailyjobs.platform == 'android', 4)],
```
jmaher: as a note, when we don't sanitize the platforms coming in, we break things like this

src/updatedb.py
```diff
@@ -123,30 +125,14 @@ def getPushLog(branch, startdate):
...
-                                 startdate.hour, startdate.minute, startdate.second,
-                                 date_xx_days_ago.year, date_xx_days_ago.month,
-                                 date_xx_days_ago.day)
-
-    db = MySQLdb.connect(host="localhost",
-                         user="root",
-                         passwd="root",
-                         db="ouija")
+    session.query(Testjobs).filter(branch == branch).\
+        filter(and_(Testjobs.date >= startdate), (Testjobs.date < date_xx_days_ago)).\
```
jmaher: probably date between would work here.

tools/dailyjobs_query.py
```diff
@@ -34,13 +39,19 @@ def parse_args(argv=None):
...
+    try:
+        session.add(dailyjob)
+    except Exception as e:
+        logger.warning(e)
+        session.rollback()
+    message = 'insert into dailyjobs (date, platform, branch, numpushes, numjobs, sumduration) ' \
+              'values ("{0}", "{1}", "{2}", {3}, {4}, {5})'.format(date, platform, branch,
+                                                                   numpushes, numjobs,
+                                                                   sumduration)
+    print message
```
jmaher: we print here, and it looks like we are not deleting the old stuff?  is there a reason to not 'delete from dailyjobs where ...' ?

tools/dailyjobs_query.py
```diff
@@ -66,14 +77,16 @@ def summarize(date, branch):
...
-
-    for rows in cur.fetchall():
+    message = "select revision,count(result),sum(duration),platform from testjobs " \
+              " where branch='{0}' and date like '{1}%' and testtype not like '%build%' " \
+              "and testtype!='valgrind' group by revision,platform;"
+    data = session.query(Testjobs.revision, func.sum(Testjobs.duration), Testjobs.platform).\
+        filter(and_(Testjobs.branch == branch, Testjobs.date.like(date),
+                    ~Testjobs.testtype.like('build'), Testjobs.testtype != 'valgrind')).group_by(
+        Testjobs.revision, Testjobs.platform).all()
+    print message
```
jmaher: we don't need to print this message, lets remove message= and print message lines.

tools/dailyjobs_query.py
```diff
@@ -66,14 +77,16 @@ def summarize(date, branch):
...
-    query = query.format(branch, date)
-    print query
-    cur.execute(query)
-
-    for rows in cur.fetchall():
+    message = "select revision,count(result),sum(duration),platform from testjobs " \
+              " where branch='{0}' and date like '{1}%' and testtype not like '%build%' " \
+              "and testtype!='valgrind' group by revision,platform;"
+    data = session.query(Testjobs.revision, func.sum(Testjobs.duration), Testjobs.platform).\
+        filter(and_(Testjobs.branch == branch, Testjobs.date.like(date),
```
jmaher: here we are missing count(result), and 'and testtype not like '%build%' " \
 +              "and testtype!='valgrind' group by revision,platform;'

jmaher: sorry, we are just missing count(result)

src/server.py
MikeLing: @jmaher yeah, I don'y know why we can get nothing by this BETWEEN. But I try this on master branch and the pure sql doesn't works well also.  I guess it's because it not only compare the daytime but also compare the detail hours and seconds.(e.g 2016-7-24 1:12:2 is bigger than 2016-7-24 so can't been filter in by "<=2016-7-24") Maybe we could enlarge the data range? (e.g we want data between 2016-7-22 and 2016-7-24, so the data range should be 2016-7-21 to 2016-7-25)

tools/dailyjobs_query.py
MikeLing: oh! I forget to delete the old things! Sorry, I just been paralysed by those sql stuff :P

src/server.py
jmaher: we should enlarge the date range- it is most important on the end date to get the latest failures.

MikeLing:
@jmaher r?

src/server.py
```diff
@@ -146,16 +149,14 @@ def run_results_day_flot_query():
...
-        cursor.execute("""select DATE(date) as day,sum(result="%s") as failures,count(*) as
-                          totals from testjobs where platform="%s" and date >= "%s" and date <= "%s"
-                          group by day""" % ('testfailed', platform, start_date, end_date))
-
-        query_results = cursor.fetchall()
+        query_results = session.query(Testjobs.date.label('day'),
+                                      func.count(Testjobs.result == 'testfailed'
+                                                 ).label("failures"),
+                                      func.count(Testjobs).label('totals')).filter(
+            and_(Testjobs.date >= start_date, Testjobs.date <= end_date)).group_by('day').all()
```
jmaher: where is 'platform' in the query?  and are we sure the DATE(date) command in mysql maps to the same format as Testjobs.date.label('day')?

src/server.py
```diff
@@ -308,13 +292,12 @@ def run_platform_query():
...
-                   where platform='%s' and buildtype='opt' and revision='%s' and
-                   build_system_type='%s' order by testtype""" % (
-            platform, cset_id, build_system_type)
-
-        cursor.execute(query)
-        test_results = cursor.fetchall()
+        test_results = session.query(Testjobs.result, Testjobs.testtype, Testjobs.date).\
+            filter(and_(Testjobs.platform == platform,
+                        Testjobs.buildtype == 'opt',
+                        Testjobs.revision == cset_summaries,
```
jmaher: why is revision == cset_summaries vs cset_id as before?

src/server.py
```diff
@@ -429,10 +406,7 @@ def run_seta_details_query():
         today = datetime.now()
         date = today.strftime("%Y-%m-%d")
 
-    db = create_db_connnection()
-    cursor = db.cursor()
-    query = "select jobtype from seta where date='%s 00:00:00'" % date
-    cursor.execute(query)
+    query = session.query(Seta.jobtype).filter(Seta.date == date).all()
```
jmaher: will this work for date?  I assume it will, but previously we had date='%s 00:00:00' to add the time stamp- is that needed?

src/server.py
```diff
@@ -506,19 +480,17 @@ def run_jobnames_query():
...
-                when 'linux' then 1 \
-                when 'osx' then 2  \
-                when 'win' then 3  \
-                when 'android' then 4 \
-                end" % (start_date, end_date)
-    cursor.execute(query)
+
+    data = session.query(Dailyjobs.date, Dailyjobs.platform, Dailyjobs.branch, Dailyjobs.numjobs,
+                         Dailyjobs.sumduration).\
+        filter(Dailyjobs.date.between(start_date, end_date)).order_by(case([
```
jmaher: here we use date between when previous queries used >= and <=, I really like date.between, and would like to use it wherever possible (even if we have to adjust the dates slightly).  Are you sure the dates are correct here, and if so, please consider doing date.between everywhere else.

tools/dailyjobs_query.py
jmaher: lets address the extra print statement

tools/dailyjobs_query.py
```diff
@@ -66,14 +80,17 @@ def summarize(date, branch):
...
-    for rows in cur.fetchall():
+    message = "select revision,count(result),sum(duration),platform from testjobs " \
+              " where branch='{0}' and date like '{1}%' and testtype not like '%build%' " \
+              "and testtype!='valgrind' group by revision,platform;"
+    data = session.query(Testjobs.revision, func.count(Testjobs.result),
+                         func.sum(Testjobs.duration), Testjobs.platform).\
+        filter(and_(Testjobs.branch == branch, Testjobs.date.like(date),
+                    ~Testjobs.testtype.like('build'), Testjobs.testtype != 'valgrind')).group_by(
+        Testjobs.revision, Testjobs.platform).all()
+    print message
```
jmaher: remove the message definition and print message lines!

tools/dailyjobs_query.py
```diff
@@ -66,14 +80,17 @@ def summarize(date, branch):
...
-    print query
-    cur.execute(query)
-
-    for rows in cur.fetchall():
+    message = "select revision,count(result),sum(duration),platform from testjobs " \
+              " where branch='{0}' and date like '{1}%' and testtype not like '%build%' " \
+              "and testtype!='valgrind' group by revision,platform;"
+    data = session.query(Testjobs.revision, func.count(Testjobs.result),
+                         func.sum(Testjobs.duration), Testjobs.platform).\
+        filter(and_(Testjobs.branch == branch, Testjobs.date.like(date),
```
jmaher: previously we had:
date like '{1}%'

that % was a wildcard to match anything that started with {1}, does:
Testjobs.date.like(date)

do the same thing?

tools/dailyjobs_query.py
MikeLing: but we print the query message before in https://github.com/mozilla/ouija/pull/109#discussion-diff-73859655L73

tools/dailyjobs_query.py
MikeLing: yep :)

tools/dailyjobs_query.py
MikeLing: alright, I forget you said we don't need it :(

src/server.py
MikeLing: it's a typo, let me fix it :P

src/server.py
MikeLing: no I don't think so....

src/server.py
MikeLing: I usr this I use this http://0.0.0.0:8157/data/setadetails/?date=2016-8-7&priority=high to test it and get:

screen shot 2016-08-08 at 8 51 59 pm

So I think it's ok :)

src/server.py
MikeLing: hmm, I'm not sure exactly and we may need test for it. But when I use http://0.0.0.0:8157/data/results/flot/day/ to test it, I got:

screen shot 2016-08-08 at 9 22 04 pm
 And it looks the same with the living version on alertmanager server http://alertmanager.allizom.org/data/results/flot/day/

MikeLing:
Hi @jmaher , I think it looks better :)

src/jobtypes.py
```diff
@@ -3,8 +3,8 @@
 import logging
 from tools.update_runnablejobs import update_runnableapi
 LOG = logging.getLogger(__name__)
-# JSONPATH = os.getcwd() + '/runnablejobs.json'
-JSONPATH = '/home/ubuntu/ouija/data/runnablejobs.json'
+JSONPATH = os.getcwd() + '/runnablejobs.json'
+# JSONPATH = '/home/ubuntu/ouija/data/runnablejobs.json'
```
jmaher: I like all the changes except this one- this will not work for alert manager- I would prefer to leave this out for now until we can configure the jsonpath better, ideally in a data/ subdir for all input data like runnable_jobs.json, preseed.json, etc...

MikeLing: oh sorry, I just need to debug on local environment and forget to change it back. Sorry :(

jmaher:
this looks great- some conflicts as I recently landed code to sanitize url params.

PR #112, opened 2016-07-28
fix the broken testtype in updatedb

MikeLing:


MikeLing:
@jmaher r?

jmaher:
thank you for this!

PR #110, opened 2016-07-27
identify the new job

MikeLing:
This PR is about to identify the new job and add it into the pressed.json.

PR #113, opened 2016-07-30
make seta return low value jobs

MikeLing:
we could make it[https://github.com/mozilla/ouija/pull/110] as option and return low value jobs as default :) @jmaher r?

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -416,6 +416,7 @@ def run_seta_details_query():
     buildbot = request.args.get("buildbot", 0)
     branch = request.args.get("branch", '')
     taskcluster = request.args.get("taskcluster", 0)
+    priority = request.args.get("priority", "low")
     jobnames = JOBSDATA.jobnames_query()
     if date == "" or date == "latest":
         today = datetime.now()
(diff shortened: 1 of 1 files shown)


src/server.py
```diff
@@ -436,8 +437,15 @@ def run_seta_details_query():
         parts = d[0].split("'")
         jobtype.append([parts[1], parts[3], parts[5]])
 
+    alljobs = JOBSDATA.jobtype_query()
+
+    # we return low value job list as default, otherwise we return high value jobs
+    if option == 'low':
+        low_value_jobs = [low_value_job for low_value_job in alljobs if
+                          low_value_job not in jobtype]
+        jobtype = low_value_jobs
```
jmaher: the comment says low value jobs by default, but I don't see how.  we should special case if option == 'high'

src/server.py
```diff
@@ -416,6 +416,7 @@ def run_seta_details_query():
     buildbot = request.args.get("buildbot", 0)
     branch = request.args.get("branch", '')
     taskcluster = request.args.get("taskcluster", 0)
+    option = request.args.get("option", "low")
```
jmaher: can we change this from 'option' to 'priority' ?

MikeLing: sure :)

src/server.py
MikeLing: We set the 'option' value as default in https://github.com/mozilla/ouija/pull/113/files/100f4fcfd6bf1544287a5261e162ff6b11bbab42#diff-94ab7c76892f0032ee9ad040e3e8f207R419. And because we store high value jobs as default in seta table, so the **if option == 'low':** make more logical sense to me :)

src/server.py
```diff
@@ -436,8 +437,15 @@ def run_seta_details_query():
         parts = d[0].split("'")
         jobtype.append([parts[1], parts[3], parts[5]])
 
+    alljobs = JOBSDATA.jobtype_query()
+
+    # we return low value job list as default, otherwise we return high value jobs
```
jmaher: please add a comment to indicate we only store high value jobs in the jobtypes

MikeLing: ok :)

PR #116, opened 2016-08-18
Ouija rewrite

MikeLing:


MikeLing:
@jmaher , I create a PR for our production end point. Due to what armenzg proposed in email, I quote

**alertmanager.allizom.org --> mozilla/ouija:production-alertmanager**
**seta-dev --> mozilla/ouija:master (Currently MikeLing's repo)**
**seta --> mozilla/ouija:heroku (branch to be created; for now master branch)**

So, we could merge this pr to mozilla/ouija:master and  mozilla/ouija:heroku. Then we could create a production version based on that :)

tools/update_runnablejobs.py
```diff
@@ -55,11 +55,16 @@ def update_runnableapi():
...
-        data = retry(requests.get, args=(url, ), kwargs={'headers': headers}).json()
-        if len(data['results']) > 0:
-            data['meta'].update({'timetamp': new_timestamp})
-            return data
+        try:
+            url = TREEHERDER_HOST.format('mozilla-inbound', task_id)
+            data = retry(requests.get, args=(url, ), kwargs={'headers': headers}).json()
+            if len(data['results']) > 0:
+                data['meta'].update({'timetamp': new_timestamp})
+                return data
```
jmaher: here you return the data, but in the except clause you do not, please move the return data line outside of the try/except block.

tools/failures.py
```diff
@@ -32,7 +32,7 @@ def get_raw_data(start_date, end_date):
 
     url = "http://alertmanager.allizom.org/data/seta/?startDate=%s&endDate=%s" % \
           (start_date.strftime("%Y-%m-%d"), end_date.strftime("%Y-%m-%d"))
-
+    print url
```
jmaher: please remove this as it is debugging only

tools/failures.py
```diff
@@ -32,7 +32,7 @@ def get_raw_data(start_date, end_date):
...
     response = retry(requests.get, args=(url, ),
```
jmaher: why are we still using an alertmanager url here?

MikeLing: because we don't have enough data on heroku for now.

src/server.py
```diff
@@ -86,11 +82,11 @@ def clean_date_params(query_dict, delta=7):
...
     start_date_param = query_dict.get('startDate') or \
-                       query_dict.get('startdate') or \
-                       query_dict.get('date')
+        query_dict.get('startdate') or \
+        query_dict.get('date')
     end_date_param = query_dict.get('endDate') or \
-                     query_dict.get('enddate') or \
-                     query_dict.get('date')
+        query_dict.get('enddate') or \
+        query_dict.get('date')
```
jmaher: do these line indentation changes pass flake8?

src/server.py
```diff
@@ -11,13 +12,15 @@
...
-import MySQLdb
 from flask import Flask, request, json, Response, abort
 
 SCRIPT_DIR = os.path.abspath(os.path.dirname(__file__))
 static_path = os.path.join(os.path.dirname(SCRIPT_DIR), "static")
 app = Flask(__name__, static_url_path="", static_folder=static_path)
 JOBSDATA = jobtypes.Treecodes()
+PORT = int(os.environ.get("PORT", 8157))
+urlparse.uses_netloc.append("postgres")
+DBURL = urlparse.urlparse(os.environ["DATABASE_URL"])
```
jmaher: can we document this better- with a comment and mention in a setup script or a readme

schduler/worker.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+import redis
+from rq import Worker, Queue, Connection
+
+listen = ['high', 'default', 'low']
+
+redis_url = os.getenv('REDISTOGO_URL', 'redis://localhost:6379')
```
jmaher: what is redis for?  why do we have a hardcoded port?

schduler/clock.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@sched.scheduled_job('interval', hours=12)
+def timed_trigger_updatedb():
+    q.enqueue(trigger_updatedb())
+
+
+@sched.scheduled_job('interval', hours=6)
+def timed_trigger_failures():
+    q.enqueue(trigger_failures())
```
jmaher: there is a slight issue here, updatedb every 12 hours is fine, but if failures.py runs at the same time we will be off- it seems that failures.py only needs to run when updatedb is completed, otherwise it isn't getting new failures will be more out of date than needed.

Does hours == time_between_runs, or does hours == time of day?

schduler/clock.py
```diff
@@ -0,0 +1,30 @@
...
+from apscheduler.schedulers.blocking import BlockingScheduler
+
+
+logger = logging.info(__name__)
+sched = BlockingScheduler()
+q = Queue(connection=conn)
+
+
+def trigger_updatedb():
+    os.system("python src/updatedb.py")
```
jmaher: updatedb.py --delta 24 --threads 4

setup.py
```diff
@@ -4,7 +4,7 @@
 required = [
     "Flask",
     "MarkupSafe",
-    "MySQL-python",
+    "psycopg2",
```
jmaher: this is fine, what about appscheduler and other stuff from requirements.txt?

src/server.py
MikeLing: yep :)

schduler/worker.py
MikeLing: these are necessary setup for reds and rq. you could find the detail in here https://devcenter.heroku.com/articles/python-rq

jmaher: can we write comments in the code and readme?

schduler/clock.py
MikeLing: @jmaher ,sorry for the late reply. I was a kind of lazy these days :P 

The 'hours' in here represent the hour between  runs. And yes you are right, these two tasks can run at the same time. Because the rq use 'worker' to execute task and it seems like rq will init a new worker for each task independently[1]. But, however, the blockingschdule we used in [2] will block other scheduler when it's running[3] and I haven't google out a solution out about the issue you mentioned above. So I propose to change the interval hours of failures.py run from **6** to **5** or **7**. I believe it can avoid the issue of two scripts been executed at the same time. :)

[1] http://python-rq.org/docs/workers/
[2] https://github.com/mozilla/ouija/pull/116/files#diff-d67473fe5745bd14193c0a193ef44dc0R5
[3] https://apscheduler.readthedocs.io/en/latest/userguide.html#choosing-the-right-scheduler-job-store-s-executor-s-and-trigger-s

setup.py
MikeLing: it's ok to not add them in requirements.txt because we add them on heroku directly thought heroku dashboard and I think people don't need scheduler when they want to debug ouija locally(they could manually run script they want to these ouija and seta :) )

MikeLing:
@jmaher r? :)

database/config.py
```diff
@@ -1,9 +1,15 @@
+import os
+import urlparse
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
-from sqlalchemy.ext.declarative import declarative_base
 
 
-engine = create_engine('mysql+mysqldb://root:root@localhost/ouija2', echo=True)
+DBURL = urlparse.urlparse(os.environ["DATABASE_URL"])
```
jmaher: if we run this locally, what value do we need in the DATABASE_URL env?  I need more information, ideally an example one and how it is set in production (if this comes by default in heroku, etc.)

MikeLing: @jmaher , hmmm so we have a little issue here. If we want to make sure people have a specific database configuration for them to test on locally, we must make a choose about which database we want to use as default.

If we wan to use **mysql**, the default configuration would be:
**database_url = 'mysql+mysqldb://root:root@localhost/ouija2'**

and if we set postgresql as default, it will become:

**database_url = 'postgresql+psycopg2://root:root@localhost/ouija2'**

NOTICE: I set database as ouija2 here is because we use ouija as our database name in alert manager server, we should switch to ouija as database name for avoiding the confusing, after the heroku server can work perfectly.

After one more thinking about this, I believe we could write a strategy to recognise the database running on the people's local environment and set database url dynamically. Like:
try:
""" If we get exception here, we believe it's running on local environment."""
os.environ["DATABASE_URL"]
except:
"""Figure out which database url we should init."""
"""But it may become pretty complex and it will make debug become harder if we have two different kinds of database."""

jmaher: please add a comment in the source code and readme about this environment variable and possible values.  We can make it smarter in the future.

database/config.py
```diff
@@ -3,13 +3,21 @@
...
+# Otherwise, the configuration is for heroku and postgresql
+try:
+    DBURL = urlparse.urlparse(os.environ["DATABASE_URL"])
+    databaseUrl = 'postgresql+psycopg2://{username}:{password}@{host}:{port}/{database}'
+    engine = create_engine(databaseUrl.format(username=DBURL.username, password=DBURL.password,
+                                              host=DBURL.hostname, port=DBURL.port,
+                                              database=DBURL.path[1:]), echo=True)
+
+except KeyError:
+    # You could set url as blow if you are using postgresql:
```
jmaher: nit: change 'blow' -> 'below'

MikeLing:
@jmaher r? :)

Issue #117, opened 2016-08-18
We need a push skipping strategy in seta


MikeLing:
we need a new push skipping strategy in seta, which is make seta return low_value_jobs and return full job list every 5 push and 90 minutes. Does it sounds good to you? @jmaher

jmaher:
this sounds great, probably a simple database table to count.  A more complete way would be a table of branch/revisions that would work in case the decision task was rerun (which happens often enough) and would give us options for multiple priorities.

Another option is a list of jobtypes:branch, with a count of requests.  If we get a series of decision task retriggers then this will be skewed.

The simplest method would be anytime the endpoint is called for a branch, increment a counter/timestamp, and when we hit our limit 5 requests or 90 minutes, then we return all high+low value jobs.  If we did this, I would like to experiment with a more complete solution- this simple solution would be faster to implement and might shed light on other quirks.

MikeLing:
Yeah, it makes sense to me! Do you think pushlog_id is a good identifier for a decision task? or head_rev?

jmaher:
I think either will work and I believe we can get both pieces of information from the decision task.  head_rev might be easier to match up to treeherder views, and to ingested data if we are trying to figure out what went wrong.  it is not as easy to read as pushlog_id, but seems like it would be slightly easier to work with.

MikeLing:
@jmaher , things seems become pretty complex after we want to calculate pushes for each branch.
I think we should exact data set from TaskRequest table and sort them by each branch. Furthermore, we should sort those data set by timestamp. So we should exact last 4 pushes for that branch and see if the priorities are all **1**(which means low jobs returned) If last 4 pushes for that branch are all **1**, we return all jobs for that push and store it in table with priority **0**.

But the problem is, we will query and sort all data sets each time when a new request comes. Maybe it's a little expensive and time consuming for a push request? What should I do to make this whole process as fast as possible? Thank you!

jmaher:
this is a great point and something I am not sure about.  In a perfect world we would have every possible job (~1500) for each branch (~10) and have a priority and history for each.  So that would be ~15000 rows in a table with:
job, branch, last_request_date, number_requests, max_requests

so every query would update the last_request_date, and increment the number_requests.  If number_requests >= max_requests, it is reset to 0.

[...]

PR #120, opened 2016-08-25
add a skipping strategy

MikeLing:


MikeLing:
it's the pr of https://github.com/mozilla/ouija/issues/117 issue.

MikeLing:
@jmaher r? :)

tools/untils.py
```diff
@@ -0,0 +1,13 @@
+"""This file defined all the objects will been used in ouiija"""
```
jmaher: 2 things:
1) comment should say SETA instead of ouija
2) lets rename this to utils.py instead of untils.py

src/server.py
```diff
@@ -78,6 +81,37 @@ def get_date_range(dates):
...
+def valve(head_rev, pushlog_id):
+    """Determine which kind of job should been returned"""
+    low_job_return = True
+    COUNTER.increase_the_counter()
+    request_list = []
+    try:
+        request_list = session.query(TaskRequests.head_rev, TaskRequests.pushlog_id,
+                                     TaskRequests.low_job_return).all()
+    except Exception:
+        session.rollback()
```
jmaher: why do we rollback() here?  possibly we should decrease_the_counter() as well?

src/server.py
```diff
@@ -78,6 +81,37 @@ def get_date_range(dates):
...
+
+    requests = {}
+    for head_rev, pushlog_id, low_job_return in request_list:
+        requests[pushlog_id] = {'head_rev': head_rev,
+                                'low_job_return': low_job_return}
+    # If this pushlog_id has been schduled, we just return
```
jmaher: s/schduled/scheduled/

src/server.py
```diff
@@ -78,6 +81,37 @@ def get_date_range(dates):
...
 
 
+def valve(head_rev, pushlog_id):
+    """Determine which kind of job should been returned"""
+    low_job_return = True
+    COUNTER.increase_the_counter()
+    request_list = []
+    try:
+        request_list = session.query(TaskRequests.head_rev, TaskRequests.pushlog_id,
+                                     TaskRequests.low_job_return).all()
```
jmaher: won't this be large over time?  can we limit it to the last 40 items?

src/server.py
```diff
@@ -78,6 +81,37 @@ def get_date_range(dates):
...
+    else:
+        # we return all jobs for every 5 pushes.
+        if COUNTER.counter >= 5:
+            COUNTER.reset()
+            low_job_return = False
+        task_request = TaskRequests(str(head_rev), str(pushlog_id), low_job_return)
+        session.add(task_request)
+        session.commit()
+
+    return low_job_return
```
jmaher: I am confused by low_job_return (and the other uses of low_value_job).  I think a longer term strategy would be to have priorities.  Ignoring the name, I am not sure what false/true means for low_job_return.  Also how low_job_return can be true.

src/server.py
MikeLing: we rollback here is because it will block other operations of this session(this session will become unusable) when an exception been raised and haven't been rollback after that. We do the same thing in updatedb.py before, but I forget to mark that related document when I read it. Please give me some time to fine that document if you need it :)

hmmm, where should we use that decrease_the_counter()? If the session rollback, we will just add that push into the database. Should we checkout the whole table again for removing those duplicate items?

src/server.py
MikeLing: yeah,and my ideal is we should maintain a fixed table size. Maybe 500 items is reasonable? We just remove those 'over size' items.

src/server.py
MikeLing: 'low_job_return' means 'we return the low job list for that push(True) or not(return all)'. I know it's a little weird in here. Could you give me some advices about it? Maybe rename it as priorities will better?

src/server.py
jmaher: I still don't understand what rollback() is for, is this the database?  if we have an exception will the data we are working with be incorrect?  My recommendation of decreasing the counter was that if we have an exception we should decrease the counter- I think our failure plan is to return ALL jobs

src/server.py
jmaher: I am not clear on what this table will hold- can you give an example of what the query will return?  I thought it was a list of revisions and what priority we assigned them- ifthat is the case and we reset the counter every 5th revision, then we don't need much data.

src/server.py
MikeLing: Ok, I got that! I will add a decrease_the_counter() and decrease the counter when we catch a exception and rollback. :)

src/server.py
jmaher: we have discussed high value vs low value for the jobs, in this case I see that:
* low_job_return == True: high value jobs
* low_job_return == False: all jobs (high value + low value)

if we think about priority as multiple levels of job values we support.  Ideally we would use a "reverse" priority system where:
priority 0 == ALL JOBS (currently low value + high value)
priority 1 == high value jobs

in the future we could have:
priority 0 == ALL JOBS
priority 1 == jobs which have found at least 1 failure in the past
priority 2 == jobs which have found at least 2 failures in the past
priority 3 == jobs which have found at least 3 failures in the past
priority 4 == jobs which have found at least 4 failures in the past

then we could determine a skipcount for each priority- as in:
every 1 push = priority 4
every 2 push = priority 3
every 3 push = priority 2
every 4 push = priority 1
every 5 push = priority 0

or more simply put- the counter that we have would start at 5 and decrement to 0, when we hit 0 reset to 5 and we just run the query of jobs from the table which match the counter :)

src/server.py
MikeLing: I'm not sure I understand the 'table will hold' in here. Here is the database table:

> id |                 head_rev                 | pushlog_id | low_job_return 
> ----+------------------------------------------+------------+----------------
>  1 | 0bd27824832d4213a90ec0ef08de422b33589358 | 80484      | t

And it will return a list of set like **[(0bd27824832d4213a90ec0ef08de422b33589358, 80484, True)]**. Does it help?

src/server.py
MikeLing: Oh great! It makes a lot of sense to me! Thank you for your help. I will address it to priority and make it match what requests above :D

src/server.py
jmaher: in this case, we have a head_rev of a push- we have <500/day, and we only care about a small range.

One thing to keep in mind- we need to track this per branch, right now: mozilla-inbound, fx-team, and autoland.  There is a possibility we need to do other branches if we extend SETA past failure discovery.  Ideally we would have this for each job to support more priorities- but I think we can make good assumptions for branch/revision.

MikeLing: ok, I could add a 'branch' column and record pushes for each branch. :)

MikeLing:
>
priority 0 == ALL JOBS
priority 1 == jobs which have found at least 1 failure in the past
priority 2 == jobs which have found at least 2 failures in the past
priority 3 == jobs which have found at least 3 failures in the past
priority 4 == jobs which have found at least 4 failures in the past
>

Maybe we need change something for this system:
**priority = None -> all jobs**

priority 0 == **all low value jobs**

these are high value jobs but just with different priority
priority 1 == jobs which have found at least 1 failure in the past
priority 2 == jobs which have found at least 2 failures in the past
priority 3 == jobs which have found at least 3 failures in the past
priority 4 == jobs which have found at least 4 failures in the past
@jmaher Do you agree about it? :)

MikeLing:
I make a mistake that I create a PR from master branch which made it's difficult to work other features from master branch. So I create another PR for it in https://github.com/mozilla/ouija/pull/133 and we could leave this pr until the other one been merged. Hope you don't mind :)

jmaher:
I think we are close or the same in thought on the priority.  My thought was to have priority a counter, it would never be None, so:
Priority 0 == ALL jobs

then we can adjust the priorities and matching of the other options.  I am fine working on #133 for a while

jmaher:
was this merged to the 'master' branch?  I am not sure where this merge went to.

jmaher:
oh, it looks as if there are merge conflicts- possibly then this will be a regular PR>

jmaher:
should we close this as well?

MikeLing:
Sure! Let's close it right now. Thank you ;)

PR #121, opened 2016-08-25
add a temp endpoint for database migration

MikeLing:


MikeLing:
@jmaher add a new(temp) endpoint and fix several flask8 errors in the repo.

src/server.py
```diff
@@ -556,6 +559,46 @@ def run_dailyjob_query():
...
+@json_response
+def run_old_job_dump():
+    limit = request.args.get('limit')
+    offset = request.args.get('offset')
+    date_format = "%04d-%02d-%02d %02d:%02d:%02d"
+    db = create_db_connnection()
+    cursor = db.cursor()
+    query = "select slave, result, build_system_type, duration, platform, buildtype, " \
+            "testtype, bugid, branch, revision, date, failure_classification, failures " \
+            "from testjobs LIMIT %s OFFSET %s" % (limit, offset)
```
jmaher: is there any concern that the query will return things in a different order making it possible to miss data on limit/offset?  

lastly how do we plan to use this, set limit to a value and keep incrementing offset until results is empty:
limit = 5000
offset = 0
results = getdump(limit, offset)
while (len(results) > 0):
    insertDataToHeroku(results)
    insertDataToHerokuStaging(results)
    offset += limit
    results = getdump(limit, offset)

jmaher:
are there other databases we care about?

src/server.py
MikeLing: Hi @jmaher, sorry for missing the comment for the endpoint.

So, the idea in here is:
we apply this endpoint in alert manager server, then we could run a simple script in heroku server and query data from alermanager server and insert them into the heroku.  And yes, the usage you mentioned above is exactly what we want. :)

However, I don't have a prefect way to ensure all the data can been migrated to heroku without any missing. Checkout and retransmission are too expensive or unnecessary I think. I think  good 'step length' can help us reduce the data missing, maybe 1000 or 500 data sets a time?

MikeLing:
AFAIK, the testjobs is only table we care about :)

src/server.py
jmaher: ok, this is good- 500 seems like a good safe route.  should we add an orderby to help ensure we get accurate data?

on the insert side, how do we ensure we don't have duplicates- I will merge this into alertmanager server, but I need to ensure releng isn't reconfig or anything like that.

PR #122, opened 2016-08-28
database migration script

MikeLing:
Hey @jmaher ,sorry for the late reply. I'm doing some preparation for beginning of school. 

I had test this script locally and it seems works good to me. But I can't deploy it on heroku directly except merge it into the master/heroku codebase. Please merge this script to master branch first and we could ask @armenzg about if our stage and production server are using same database on heroku.

### DIFF
--- a/tools/database_migration.py
+++ b/tools/database_migration.py
@@ -1,0 +1,45 @@
+import argparse
+import requests
+from redo import retry
+from database.config import session
+from database.models import Testjobs
+
+# alertmanager server URL
+URL = "http://alertmanager.allizom.org/data/dump/?limit=%d&offset=%d"
+
+
+def migration(args):
+    limit = int(args.limit)
+    offset = 0
+    url = URL % (limit, offset)
+    response = retry(requests.get, args=(url, )).json()
+    datasets = response['result']
+    while len(datasets) > 0:
+        for data in datasets:
+            Testjob = Testjobs(data['slave'], data['result'], data['build_system_type'],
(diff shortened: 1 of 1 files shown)


tools/database_migration.py
```diff
@@ -0,0 +1,45 @@
...
+                               data['failure_classification'], data['failures'])
+            try:
+                session.add(Testjob)
+                session.commit()
+
+            except:
+                session.rollback()
+
+            finally:
+                session.close()
```
jmaher: do we need to close the session each time, or can we leave it open until the while loop is finished?

tools/database_migration.py
```diff
@@ -0,0 +1,45 @@
...
+            Testjob = Testjobs(data['slave'], data['result'], data['build_system_type'],
+                               data['duration'], data['platform'], data['buildtype'], data['testtype'],
+                               data['bugid'], data['branch'], data['revision'], data['date'],
+                               data['failure_classification'], data['failures'])
+            try:
+                session.add(Testjob)
+                session.commit()
+
+            except:
+                session.rollback()
```
jmaher: there is no indication that there will be missed data- we need a way to log our progress.

jmaher:
please address my two issues here, then we can get this on seta-dev first, run the script, if it works- we can determine if the DB is the same- It would be nice if they had the same backend data for testjobs, the other data might not be good to share if we change failures.py, etc. it could cause confusion.

MikeLing:
@jmaher r?

jmaher:
merged this- now we need to run this on the heroku commandline?

PR #133, opened 2016-09-04
add a skipping strategy

MikeLing:


MikeLing:
@jmaher f? I haven't test it, just want to make sure it's on the right direction :)

src/server.py
```diff
@@ -148,6 +156,38 @@ def binify(bins, data):
...
 
 
+def valve(head_rev, pushlog_id, branch, priority):
+    """Determine which kind of job should been returned"""
+    priority = priority
+    BRANCH_COUNTER.increase_the_counter(branch)
+    request_list = []
+    try:
+        request_list = session.query(TaskRequests.head_rev, TaskRequests.pushlog_id,
+                                     TaskRequests.priority).limit(40)
```
jmaher: why the limit 40?

src/server.py
```diff
@@ -148,6 +156,38 @@ def binify(bins, data):
     return result
 
 
+def valve(head_rev, pushlog_id, branch, priority):
+    """Determine which kind of job should been returned"""
+    priority = priority
+    BRANCH_COUNTER.increase_the_counter(branch)
```
jmaher: we should make sure that 'branch' is valid- we will get called for EVERY decision task.  One way to reduce this is to make sure that we hardcode the list of SETA branches in the in-tree code for the decision task where we call SETA server.  But still, this needs to be 100% robust.

src/server.py
jmaher: could we just limit this to pushlog_id, then we don't need the list?

src/server.py
```diff
@@ -148,6 +156,38 @@ def binify(bins, data):
...
+        priority = requests.get(pushlog_id)['priority']
+    else:
+
+        # we return all jobs for every 5 pushes.
+        if RequestCounter.BRANCH_COUNTER[branch] >= 5:
+            RequestCounter.reset(branch)
+            priority = None
+        task_request = TaskRequests(str(head_rev), str(pushlog_id), priority)
+        session.add(task_request)
+        session.commit()
```
jmaher: I do question if we need RequestCounter and TaskRequests- I would believe we can get by just fine with TaskRequests since pushlog_id will be branch specific- lets add a field 'branch' to the TaskRequests, and we should be able to simplify.  Now if we request data by pushlog_id, then it won't work- I think we need to think about query speed and determine if two small queries is better than 1 sort of small query.  Also for code maintainability- overall this two table structure I don't like, but it might be most efficient if done right.

src/server.py
MikeLing: we make it 40 because I remember you said we need to do so in https://github.com/mozilla/ouija/pull/120#discussion_r76230891

MikeLing:
So, which priority will be used to represent low value jobs if **0** represent ALL jobs

jmaher:
we would never just run low-value jobs, it would be all jobs- right now either:
* high value only
* all jobs (high+low value)

with some future work to seta we can then split high value jobs into different priorities.

MikeLing:
@jmaher hmmm, sorry. I still confuse the idea in https://github.com/mozilla/ouija/issues/117#issuecomment-242820683. Let me explain my understanding to you and ask some feedback.

## assumming it's a 5 priorities system

> the table we are going to define

Skipcounter: including  **counter**, **branch**

> the variable we are going to define

REQUESTLIST: to cache the push we had returned jobs for, which including **head_rev**, **push log_id**, **priority**. Sort of {pushlog_id1[or head_rev]: priority, pushlog_id2: priority, pushlog_id3: priority}
**maybe defined as a deque and maxlength limit as 50 items**

> the whole process here


```
A request comes in with _head_rev, _pushlog_id, _branch
if it's request from taskcluster:
    priority = valve()
if priority is not 0:
    jobs = query(select job from seta where priority=priority)
else:
    jobs = all jobs # which means return all jobs

def valve():
    priority = query(select priority from Skipcounter where branch=_branch )
    if pushlog_id in REQUESTLIST.keys():
        return REQUESTLIST[pushlog_id]
        priority += 1
    """It's possible that priority greater than 5 because maybe it's the 5th request and the push had  been cached in  REQUESTLIST which we will plus one for it."""
    if priority >= 5:
        priority = 0
    else priority += 1
    update priority for Skipcounter where branch=_branch NOTE: maybe we could add a column **next priority** with a sql logic which always bigger than priority column and reset to zero if it's greater than 5
    return priority
```

jmaher:
this is pretty close to my thoughts.  My original thoughts were to keep track of the revisions we ran tests queries on, I believe that is not needed anymore and we just need a branch/count- but possibly there is a use case I am not thinking about.

The only other change is that I would like to have a single list of all jobs with a priority assigned to each job.  This single list of jobs would then be cumulative-  for example, the highest value jobs should be run on every commit, whereas the lowest priority jobs will be run on every 5th commit.  Therefore anything with a priority between 0 and 5 we would want to ensure we add more jobs on each test run, making this:

```
if priority is not 0:
    jobs = query(select job from seta where priority<priority)
else:
    jobs = all jobs # which means return all jobs
```

doing this, we would reset the priority=5 when priority <=0, decrement it each call.

The only thing missing here is if we decide that some branches/platforms/jobs should run every 7th or 10th push, how would we change this?  Possibly we need to have a table of jobs with:
name, priority, skipcount

that would assume the the jobname table contained platforms in the job names.

Lets first make sure we both understand the cumulative set of jobs based on priority and ensure we think about all the cases and agree upon which tables we need.  I am referring to the REQUESTLIST table- can you outline the reasons for needing it? (I know I suggested it originally)

jmaher:
The REQUESTLIST table would be useful for a few things:
1) making sure we ignore retriggered decision tasks
2) dealing with some race conditions, or if there were errors somewhere in the system
3) looking back historically at push rates, and what SETA was doing

the second case is the most interesting one.

Thinking about custom scheduling for different scenarios (like PGO, Talos, a certain platform like WindowsXP) we would need more granularity than we are proposing in both the REQUESTLIST and the SKIPCOUNT tables.  I am thinking we will end up with a jobpriority table that will list all jobs base on platform (probably not branch) with defined skipcount and priority information.  Effectively this would be the output of failures.py.  Right now failures.py output all HIGH VALUE JOBS only into a table- one set of jobs for each day it runs.  This is redundant, but I see it instead:
1) updating the priority values of jobs in the jobpriority table
2) updating the SETA table with what jobs changed priorities so we have historical data

If we wanted to support a variable amount of skipcount data we could adjust to a model where we always increment the skipcounter and never reset it.

then we might end up with logic like this:
...
branch_priority = X # incremented skipcount value
max_priorities = Y # highest value priority from jobpriority table (current SETA is 1, our discussions is 5) 
priorities = [p for p in max_priorities where p % branch_priority == 0] # find all priorities that are ready to be scheduled

# ideally we would have the jobpriority table cached in memory and avoid the query
jobs = query(select jobs, platform from jobpriority where priority in priorities) # assuming we can do a query like this and it is efficient
...

by using the % operator we can always run every time, or every 8th time with no problems.

I believe this covers the skipping by skipcount, but we are ignoring timestamps.  I think when we update the counter, we update a timestamp as well- we could do similar logic to the above psuedo code to account for a timestamp > max_time.

MikeLing:
@jmaher I'm still confusing about the usage of jobpriority table and how we use it. Could you describe   how this table should looks like? ASFAIK, I guess it may looks like(if it's about branch not platform):

  branch  | priority0 | priority1 | ......
  ------------- | ------------- | ------------- | -------------
  mozilla-inbound  | [job1,job2,job3....] | [job4,job7,job9....] | ......
  mozilla-central  | [job3,job4,job6....] | [job1,job2,job5....] | ......
  ...... | ...... | ......

jmaher:
I was thinking more like this:
  jobname  | platform | priority | skipcount | timeout
  ------------- | ------------- | ------------- | ------------- | -------------
  mochitest-1 | linux64 | 1 | 1 | 0
  mochitest-1 | win32 | 5 | 5 | 5400
  mochitest-1 | osx10.10 | 5 | 5 | 5400
  mochitest-1 | win64 | 5 | 5 | 5400
  mochitest-1 | android | 5 | 5 | 5400
  cppunit | win64 | 1 | 1 | 0
  cppunit | win32 | 10 | 10 | 7200
  ...... | ...... | ...... | ...... | ......

this would be a long table, where we have about between 200 unique test jobs and 15 platforms- so that would be a table of 3000 jobs to maintain.

In addition to the above mentioned JOBPRIORITY table, we could have the BRANCHCOUNT table:
  branch  | count | datetime
  ------------- | ------------- | ------------- 
   mozilla-inbound | 45 | 2016-09-05 12:13:14
   autoland | 42 | 2016-09-05 10:11:12
  ... | ... | ...

I think those two tables will help- notice how I keep changing my thoughts here- so keep asking questions and offering suggestions- we will have a solution before too long.

MikeLing:
what's the different between skipcounter and priority? And what's the timeout for?

jmaher:
i think skipcount would be the same as priority- so we could merge the fields!  

timeout is for when we have 1 push every 4 hours on a holiday- then we would want to run all tests- so the timeout is by default 90 minutes (5400 seconds)

MikeLing:
Maybe we should rename timeout as interval ? And is that mean we no longer need seta table but use job priority table to instead of it when we need to return job list? Maybe we should merge them(seta table and jobpriority table ) in the future?

jmaher:
agreed- lets remove priority and keep skipcount and interval :)

I also agree that SETA table is redundant now- we might as well go with our new architecture and stop using the SETA table sooner rather than later.

jmaher:
@MikeLing would it be reasonable to get a version of this ready to land on the master branch this week or weekend?  possibly next week I can have it deployed and tweak it until it works.  I believe the data migration is done- I will know in 2 days when a few automated updates are done.

MikeLing:
@jmaher This patch can works good after resolved the conflicts in here, I think. But I haven't address it as you mentioned above. 

We need to add two more tables "Jobpriority" and "Branchcounter". And I think we need to rewrite some parts of code in seta.py and failures.py, due to we are going to use Jobpriority table right now. Then we need to make these things works good for job counting and job priority changing. At last, the new endpoint must return all jobs as we expect.

If you like it, we could resolve the conflict in here and to see what we can got. Then we could comprehensive the whole strategy in here.

jmaher:
yes, lets fix the conflicts and we can follow up on remaining items!

MikeLing:
@jmaher as I mentioned on irc, I had fix the conflicts in here. But I need some more time to make it comprehensive :)
r?

jmaher:
lets get this in for now, I will see how this works and we will continue to work on the priority based system for all job types.

MikeLing:
I would like to close this PR due to https://github.com/mozilla/ouija/pull/133 and https://github.com/mozilla/ouija/pull/141 been closed :)

jmaher:
agreed

PR #136, opened 2016-09-07
Add try and exception

MikeLing:


tools/failures.py
```diff
@@ -31,9 +32,12 @@ def get_raw_data(start_date, end_date):
...
           (start_date.strftime("%Y-%m-%d"), end_date.strftime("%Y-%m-%d"))
-
-    response = retry(requests.get, args=(url, ),
-                     kwargs={'headers': headers, 'verify': True})
+    try:
+        response = retry(requests.get, args=(url, ),
+                         kwargs={'headers': headers, 'verify': True})
+    except Exception as e:
+        logger.debug("The request to %s is failed due to %s" % (url, e))
+        return
```
jmaher: can we return nothing here and all code that calls this method will be fine?  should we return a  blank object {} ?

MikeLing: alright, I think I understand what's your idea in here. We should keep seta running even something wrong on the network, right? I think return a {} is a good idea and I will look after the other places :)

jmaher: correct, lets see where else we can make things more robust!

MikeLing:
hmm, it seems like too many conflicts in here, due to I compare two different branch(**master** and **alert manager** ). Please let me close this one directly :)

MikeLing:
https://github.com/mozilla/ouija/pull/139

PR #141, opened 2016-09-19
Add jobpriorities table, store runnablejobs and preseed data there wi…

jmaher:
…th priority, timeout, [expiration].

This works like this:
1) when server.py is started, we reparse preseed.json (it only changes with code updates)
2) when failures.py is ran, we reference all data from jobpriorities table
3) whenever update_runnablejson.py is called to download, we update jobpriorities table (either in server.py or failures.py)

This changes a few things:
1) preseed.json is merged into jobpriorities table- an expires value of NULL means no preseed data
2) we don't need SETA table anymore- we lose the ability to track changes/day, but we get a smaller table that supports the needs for taskcluster

Still to do:
1) a lot of testing
2) write a lot of tests
3) remove all references to SETA data
4) consider adding date_changed field to jobs table (for new jobs, or ones where priority is changed)
5) run failures.py in multiple passes to assign priorities
6) hook up to valve() and create branchcount table

@MikeLing f?


MikeLing:
Hey @jmaher , it's a really fantastic work! I saw a lot of code been improved and many logic become clear to me. Thank you Joel :)

f+

tools/failures.py
```diff
@@ -90,6 +95,57 @@ def insert_in_database(to_insert, date=None):
...
+        if dv.date() <= now.date():
+            conn = engine.connect()
+            statement = update(JobPriorities)\
+                          .where(JobPriorities.id == item[1])\
+                          .values(expires=None)
+            conn.execute(statement)
+
+
+def update_jobpriorities(to_insert, _priority, _timeout):
+    # to_insert is currently high priority, pri=1 jobs, all else are pri=5 jobs
```
MikeLing: maybe we could make  to_insert looks like as a dict including job name and job priority (e.g   _**{job: job priority}**_ ), after we have 5 priorities job system.

jmaher:
I did more testing, and then added buildsystem so we can deal with buildbot properly as well as taskcluster.  This found some errors I was having with duplicate jobs in the jobpriorities table.  Much more testing also yielded an issue where priorities were not being updated after failures.py was ran.

I feel there is more testing to do, but more confident in the changes.

what remains is:
1) a bit more testing
2) write a lot of tests
3) remove remaining references to SETA data (database tables- we don't use them now) and the server.py api interfaces
4) run failures.py in multiple passes to assign priorities
5) hook up to valve() and create branchcount table

I am not interested in:
1) consider adding date_changed field to jobs table (for new jobs, or ones where priority is changed) - I vote not to, it isn't that critical and we can add it later if we feel that data is useful.

jmaher:
and adding more code, this is not fully complete, but it does leave room for testing more.

I need to do more testing for the taskcluster api return values as well as really beating up the api for buildbot.

of course other todo items are:
1) redoing the taskrequest table as per todo comments
2) simplifying the priority status and finding ways to query smoother
3) finish hooking up to valve()

@MikeLing if you have some time could you look over this code again, it is getting closer to complete- if you see anything you would like to change call it out!  I do like the idea of to_insert switching to having a priority- That seems like the next project after we get priority 1 and 5 working for taskcluster.

src/server.py
```diff
@@ -433,50 +434,38 @@ def run_seta_query():
...
-
-    return {'dates': retVal}
-
-
 @app.route("/data/setadetails/")
 @json_response
 def run_seta_details_query():
-    startDate, date = clean_date_params(request.args)
-    active = sanitize_bool(request.args.get("active", 0))
+    inactive = sanitize_bool(request.args.get("inactive", 0))
```
MikeLing: why we need this inactive? Just use **priority** is not enough?

src/server.py
```diff
@@ -433,50 +434,38 @@ def run_seta_query():
...
-    date = "%s" % date
-    query = session.query(Seta.jobtype).filter(Seta.date == date).all()
+    date = str(datetime.now().date())
+
+    if inactive == 1:
+        priority = 5
+    else:
+        priority = 1
+
+    # TODO: we can make this a variable priority in the future based on input
```
MikeLing: could you explain this idea more detail ? Just curious and want to know more about it. Thank you :)

src/server.py
```diff
@@ -487,33 +476,26 @@ def run_seta_details_query():
...
     if priority == 0:
+        jobtype = alljobs
+    # priority =5 run all low value jobs
+    elif priority == 5:
         low_value_jobs = [low_value_job for low_value_job in alljobs if
                           low_value_job not in jobtype]
         jobtype = low_value_jobs
+    # priority =1, run all high value jobs
+    elif priority == 1:
+        pass # use jobtype as a high value query
```
MikeLing: so, we still make it return **high** and **low** value job? I think we are going to return a job list depend on priority that 

0 == all jobs
1 == one failure related job
2 == two failure related job 
3 == three failure related job
4 == four failure related job

which means we don't need to return low value job anymore.

src/server.py
```diff
@@ -605,5 +588,105 @@ def template(filename):
...
+
+    # Preseed data will have fields: buildtype,testtype,platform,priority,timeout,expires
+    # The expires field defaults to 2 weeks on a new job in the database
+    # Expires field has a date "YYYY-MM-DD", but can have "*" to indicate never
+    # Typical priority will be 1, but if we want to force coalescing we can do that
+    # One hack is that if we have a * in a buildtype,testtype,platform field, then
+    # we assume it is for all flavors of the * field: i.e. linux64,pgo,* - all tests
+    # assumption - preseed fields are sanitized already - move parse_testtype to utils.py ?
+    for job in preseed:
+        _buildsystem = "buildbot" # use this as a default
```
MikeLing: could you tell me the usage of ** * ** ? And what will happened when the build system is "taskcluster".

src/server.py
jmaher: oh!  you are right.  This is only used for supporting buildbot legacy api!  I will add a comment.

src/server.py
jmaher: Right now we are querying based on priority 1 (high value jobs).  Then we either return that, or the entire list of jobs - priority 1 (low value jobs).

When we have other priorities, I imagine us querying all jobs into the scope and filtering based on the input.  Maybe where priority in [1,2,3] or something like that.

src/server.py
jmaher: I was thinking it would be different:
1 == detects 4 or more failures
2 == detects 3 failures
3 == detects 2 failures
4 == only 1 failure
5 == all jobs

right now that maps to:
1 == high value jobs
5 == low value jobs + high value jobs

this needs to be more robust and documented.  The ideal plan is to switch to this model:
select counter from taskrequests where branch = branch;
jobs_to_run = runnable_jobs.filter(priority=1)
if counter % 2 == 0:
   jobs_to_run += runnable_jobs.filter(priority=2)
if counter % 3 == 0:
   jobs_to_run += runnable_jobs.filter(priority=4)
if counter % 4 == 0:
   jobs_to_run += runnable_jobs.filter(priority=4)
if counter % 5 == 0:
   jobs_to_run += runnable_jobs.filter(priority=5)

this would then mean we don't run all the tests at once.  This also ignores the timeout clause- I am sure we can get there.

src/server.py
jmaher: I decided that '*' would represent all.  This could be problematic if we have a third system, but right now we have just 2: buildbot and taskcluster.

You ask a good question about why this is defaulted to "buildbot" and what happens with taskcluster jobs.  I had to re-read the code a few times to get this.

In this function we are specifically dealing with new preseed data.  Right now preseed is primarily talos which is all buildbot.  So the idea is we would add a new job that doesn't exist in with a default buildsystem.  If the job exists already, then we compare the data and make sure that we support both buildsystems with a "*".

I have a bug in here, and I believe in preseed.json we need to add "buildsystem" to the fields.  This would allow us to not hardcode data and make things much more accurate!

Thanks for pointing this out.

PR #150, opened 2016-09-26
update taskcluster endpoint

MikeLing:


MikeLing:
@jmaher f?

src/server.py
```diff
@@ -437,25 +405,13 @@ def run_seta_query():
 @app.route("/data/setadetails/")
 @json_response
 def run_seta_details_query():
-    # TODO: remove inactive when buildbot api queries s/inactive/priority/
-    inactive = sanitize_bool(request.args.get("inactive", 0))
```
jmaher: I would like to leave a comment in here as this would break buildbot if we merged this or switched buildbot.  I am fine removing it though :)

src/server.py
```diff
@@ -437,25 +405,13 @@ def run_seta_query():
...
     buildbot = sanitize_bool(request.args.get("buildbot", 0))
     branch = sanitize_string(request.args.get("branch", ''))
     taskcluster = sanitize_bool(request.args.get("taskcluster", 0))
     priority = int(sanitize_string(request.args.get("priority", '1')))
-
+    jobtype = []
```
jmaher: don't we set this a few lines below?  we should remove one of them.

src/server.py
```diff
@@ -464,37 +420,97 @@ def run_seta_details_query():
...
+            branch_data = TaskRequests(str(branch), 1, str(time_of_request), 5400)
+            try:
+                session.add(branch_data)
+                session.commit()
+            except Exception:
+                session.rollback()
+
+            finally:
+                session.close()
+            counter, string_time, reset_delta = 1, time_of_request, 5400
```
jmaher: I don't like this style of assignment, it is a neat shortcut, but potential for confusion.  Also 5400 is 90 minutes (not seconds), which looks correct- but lets make it a constant, even if it is in the function.

src/server.py
```diff
@@ -464,37 +420,97 @@ def run_seta_details_query():
...
         else:
-            priority = 0
+            counter, string_time, reset_delta = branch_info[0]
+            conn = engine.connect()
+            statement = update(TaskRequests).where(
+                TaskRequests.branch == branch).value(
+                counter == counter + 1)
+            conn.execute(statement)
+
+        last_update_time = datetime.datetime.strptime(string_time, "%Y-%m-%d %H:%M:%S")
```
jmaher: string_time is only set in the case where the branch doesn't exist.

src/server.py
```diff
@@ -464,37 +420,97 @@ def run_seta_details_query():
...
+        delta = (time_of_now - last_update_time).total_seconds()
+
+        # we should update the time recorder if the elapse time had
+        # reach the time limit of that branch.
+        if delta >= reset_delta:
+            conn = engine.connect()
+            statement = update(TaskRequests).where(
+                TaskRequests.branch == branch).value(
+                datetime == datetime)
+            conn.execute(statement)
```
jmaher: I would like to update the database only once here- to reduce operations.  We now have:
query for jobs
query for branch data
insert branch data (rare)
increment counter
update timestamp

I would like to increment counter and update timestamp at the same time- we should update the timestamp for all calls, the reset would force us to run all jobs instead of use the specific priority.

jmaher: a new thought- if we have long time with no tests (90 minutes) then we want to run all jobs- not just whatever matches the priority- this should be documented better with comments, it isn't obvious and not the best design on my part.

src/server.py
```diff
@@ -464,37 +420,97 @@ def run_seta_details_query():
...
+
+        for d in query:
+            # we only return that job if it hasn't reach the timeout limit.
+            if delta < d[4]:
+                # Due to the priority of all high value jobs is 1, and we
+                # need to return all jobs for every 5 pushes(for now).
+                if counter + 1 % d[3] != 0:
```
jmaher: if we have a variable counter and increment it originally- we can add that counter to the database and just use it here  without (counter +1).

src/server.py
```diff
@@ -464,37 +420,97 @@ def run_seta_details_query():
...
     # priority =5 run all low value jobs
-    elif priority == 5:
-        low_value_jobs = [low_value_job for low_value_job in alljobs if
-                          low_value_job not in jobtype]
-        jobtype = low_value_jobs
-    # priority =1, run all high value jobs
-    elif priority == 1:
-        pass # use jobtype as a high value query
+    else:
+        joblist = [job for job in query if job[3] == priority]
```
jmaher: my thoughts here are we could do;
for p in range(1,5):
    if p % counter == 0:
        priorities.append(p)

joblist = [job for job in query if job[3] in priorities]

^ this will allow us to match multiple priorities easily (and take care of the high/low value case)

src/server.py
MikeLing: About this, after we remove inactive variable, we should change the way which buildbot make request. The buildbot should use priority instead of inactive.

src/server.py
MikeLing: Ok, I got it!

src/server.py
jmaher: we will have to port some changes to the alertmanager server first- can you file an issue for Ouija so we can track this?

src/server.py
MikeLing: string_time is the time of that branch been returned as full job list last time. I think the time_of_request is about the case where the branch doesn't exist.

src/server.py
MikeLing: > if p % counter == 0:

Do you mean **counter % p == 0**? Because counter is bigger than p and if p % counter will alway equal to itself.

jmaher:
Please double check the variables are initialized

On Sep 26, 2016 04:50, "Tiramisu 1993" <notifications@github.com> wrote:

*@MikeLing* commented on this pull request.
------------------------------

In src/server.py <https://github.com/mozilla/ouija/pull/150>:

>          else:
-            priority = 0
+            counter, string_time, reset_delta = branch_info[0]
+            conn = engine.connect()
+            statement = update(TaskRequests).where(
+                TaskRequests.branch == branch).value(
+                counter == counter + 1)
+            conn.execute(statement)
+
+        last_update_time = datetime.datetime.strptime(string_time,
"%Y-%m-%d %H:%M:%S")

string_time is the time of that branch been returned as full job list last
time. I think the time_of_request is about the case where the branch
doesn't exist.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mozilla/ouija/pull/150>, or mute the thread

jmaher:
Yes, I have it backwards.  Thanks for catching my error.

src/server.py
MikeLing: It's a good idea! But I think we could implement it after we have multiple priorities system(right now we only have 2 priorities).

jmaher:
Yes, I agree.  We can support it later :)

src/server.py
jmaher: please leave this in or disable it with a # instead of deleting the line.

jmaher: wait, I see the issue opened up- lets merge this code and track it with the issues.

PR #154, opened 2016-10-03
Fix the error in setadetail endpoint

MikeLing:


jmaher:
the comment here doesn't match up with the change.   I have no problem with the change.  can you --amend your commit and adjust the comment?

PR #155, opened 2016-10-03
Do not drop jobpriorities as it might not exist

armenzg:
r? @jmaher

That drop statement was getting on the way when starting from a clean state.


jmaher:
thanks for the cleanup!

PR #156, opened 2016-10-03
Create only one schema definition

armenzg:
Instead of carrying a schema per table I believe it is easier to use
only one sql file for various reasons.

It is exactly the same as what mysqldump would use:
 * mysqldump -u root --no-data ouija

It removes the wondering if there's a reason for them to be separate
(e.g. some apply to ouija while others to seta).


jmaher:
there are conflicts with this, can you fix those and we can land this?

PR #157, opened 2016-10-03
Add troubleshooting section + reword some sentences

armenzg:


jmaher:
thanks, this is a good step in the right direction.

PR #158, opened 2016-10-06
Refactor SETA's code to update job's priority table + DB changes

armenzg:
Hi @jmaher @MikeLing,
This PR is trying to help me understand the code.
I'm doing some small refactors as well as asking questions.

Please let me know if there are any pieces of the code that you would like me to aim to land.
For the most part I'm more interested on answers to my questions rather than landing the code.
Refactoring only helps me understand it better.

I will be back on Tuesday.


tools/failures.py
```diff
@@ -182,7 +182,79 @@ def analyze_failures(start_date, end_date, testmode, ignore_failure, method):
...
+    #	   "job_group_symbol": "tc-R",
+    #	   "job_type_description": "Reftest run",
+    #	   "job_type_name": "desktop-test-linux64/opt-reftest-8",
+    #	   "job_type_symbol": "R8",
+    #	   "platform": "linux64",
+    #	   "platform_option": "opt",
+    #	   "ref_data_name": "desktop-test-linux64/opt-reftest-8",
+    #	   "result": "runnable",
+    #	   "state": "runnable"
+    #	 },
```
jmaher: this data is not relevant to failures.py- failures.py only care about buildtype,testtype,platform from a sql query.

tools/update_runnablejobs.py
```diff
@@ -47,12 +47,13 @@ def update_runnableapi():
     # The format of expires is like 2017-07-04T22:13:23.248Z and we only want 2017-07-04 part
     expires = latest_task['expires'].split('T')[0]
     time_tuple = datetime.datetime.strptime(expires, "%Y-%m-%d").timetuple()
-    new_timestamp = time.mktime(time_tuple)
+    new_timestamp = time.mktime(time_tuple) # e.g. 1499140800.0 seconds since the epoch
```
jmaher: that is useful to add in!  thanks:)

tools/failures.py
```diff
@@ -182,7 +182,79 @@ def analyze_failures(start_date, end_date, testmode, ignore_failure, method):
...
+    # It also calls add_jobs_to_jobpriority()
     update_runnableapi()
+    import sys
+    sys.exit(1)
```
jmaher: we would need to remove the sys.exit(1)

tools/update_runnablejobs.py
```diff
@@ -47,12 +47,13 @@ def update_runnableapi():
...
     path = get_rootdir() + '/runnablejobs.json'
 
     # we do nothing if the timestamp of runablejobs.json is equal with the latest task
     # otherwise we download and update it
     if os.path.isfile(path):
+        # XXX: this is inefficient as json loading such big file is a slow method
```
jmaher: is runnable jobs that large?  would there be a better way to support getting all the valid job types?

tools/update_runnablejobs.py
```diff
@@ -62,19 +63,25 @@ def update_runnableapi():
...
             data = query_the_runnablejobs(new_timestamp, task_id)
-            new_jobs = add_jobs_to_jobpriority(new_data=data,
-                                               priority=1,
-                                               timeout=0,
-                                               set_expired=True)
+            # We're calling with set_expired to True since we expect the new
+            # runabble file might have new jobs
+            # XXX: How do we know that in this case we're going to have new jobs instead of
+            #      the other case below? Does the timestamp in meta that runnable apis give us
+            #      mean that it actually has new data?
```
jmaher: here we assume that if the file exists that it was a valid file and we are not bootstrapping the system.  this is an assumption and has potential to break- thanks for raising the concern- please don't comment this out though.

tools/update_runnablejobs.py
```diff
@@ -62,19 +63,25 @@ def update_runnableapi():
...
     else:
         print "It's going to help you download the runnable jobs file."
         data = query_the_runnablejobs(new_timestamp, task_id)
         if data:
             with open(path, 'w') as f:
-                json.dump(data, f)
-            add_jobs_to_jobpriority(new_data=data, priority=5, timeout=5400)
+                json.dump(data, f, indent=2, sort_keys=True)
+            # We're calling this without setting set_expired to True
+            #add_jobs_to_jobpriority(new_data=data, priority=5, timeout=5400)
```
jmaher: this is by design- we do not have a pre-existing file and we assume we are bootstrapping the system.

tools/update_runnablejobs.py
```diff
@@ -144,44 +151,15 @@ def parse_platform(platform):
     return platform
 
 
-def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=False):
+def _update_database_with_new_jobs(new_data, set_expiration=False):
+    ''' Return jobs that are new.'''
+    # XXX: jmaher: we might want to change how this method is implemented to do directly query the
+    #              DB by creating a good index. It is probably faster than iterating over rows
```
jmaher: I found that doing hundreds of queries were not that efficient on heroku- in fact I spent about 4 hours trying to get operations to complete before timeout due to queries in a loop.  Possibly you have better ideas for more efficient queries or smarter ones.

tools/update_runnablejobs.py
```diff
@@ -144,44 +151,15 @@ def parse_platform(platform):
...
-               job['platform'] == nd['platform'] and \
-               job['platform_option'] == nd['platform_option']:
-                found = True
-                break
-
-        if found:
-            if job['build_system_type'] != nd['build_system_type']:
-                job['build_system_type'] = '*'
-        no_duplicates.append(job)
-    new_data['results'] = no_duplicates
```
jmaher: the duplicate removal is pretty key to quicker performance when dealing with the database

tools/update_runnablejobs.py
```diff
@@ -193,16 +171,20 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
 
     # Now loop through the reduced jobs and update the database as needed
     for job in new_data['results']:
+        # XXX: from earlier in the code, I believe we could have two jobs one with the normal
+        # buildsystem type and the other with *
         _buildsystem = job["build_system_type"]
```
jmaher: we cannot have duplicates in our jobpriorities table- we either need no * and to support explicitly 'buildbot' and 'taskcluster' with about 800 duplicated jobs.  That makes it very difficult to support preseed.json, but it could be done.  The only case I see for not doing '*' is if we expect different failure patterns on taskcluster vs buildbot when we run side by side.  By failures, I am not referring to intermittent, but real failures that cause a backout or hotfix.

tools/update_runnablejobs.py
```diff
@@ -193,16 +171,20 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
         platform = job["platform"]
         testtype = job["testtype"]
 
         found = False
         found_id = None
+        # XXX: jmaher do we have to iterate through every single row? or can we break early?
```
jmaher: good catch, we should break when found=True

tools/update_runnablejobs.py
```diff
@@ -213,7 +195,8 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
         # We have new jobs from runnablejobs to add to our master list
         if not found:
             _expired = None
-            if set_expired:
+            if set_expiration:
+                # XXX: I wonder if we could set in here the default values (e.g. priority=1 et al)
```
jmaher: yes, we need to default to p=1, t=0;

tools/update_runnablejobs.py
```diff
@@ -243,6 +226,49 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
+def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=False):
+    if not new_data:
+        return
+
+    # This is a loop to remove duplicates (including buildsystem -> * if needed)
+    # By doing this it allows us to have a single database query
+    # no_duplicates will contain a subset which excludes:
+    #   * jobs that don't specify the build_platform
+    #   * jobs that don't specify the testtype
+    #   * XXX: unclear why we append the job again but with *
```
jmaher: we do not append the job with *, instead we replace the buildsystem type with '*' if we find an identical job between both systems.

tools/update_runnablejobs.py
```diff
@@ -243,6 +226,49 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
                           .values(buildsystem=_buildsystem)
             conn.execute(statement)
 
+
+def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=False):
```
jmaher: you do not use priority or timeout here, they should be proxied to the next function

tools/update_runnablejobs.py
```diff
@@ -243,6 +226,49 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
+                break
+
+        no_duplicates.append(job)
+
+    # XXX: Does this affect the new_data outside of this function?
+    new_data['results'] = no_duplicates
+
+    added_jobs = _update_database_with_new_jobs(new_data=new_data)
+
+    # XXX: We don't actually use this information outside of the function
```
jmaher: correct- possibly we don't need to return it- I had returned the information for debugging purposes while developing it.

tools/update_runnablejobs.py
armenzg: nvm me. It's only 2mb.
```bash
armenzg@armenzg-thinkpad:~/repos/ouija$ ls -l --block-size=M ~/.mozilla/seta/runnablejobs.json 
-rw-rw-r-- 1 armenzg armenzg 2M Oct  6 16:45 /home/armenzg/.mozilla/seta/runnablejobs.json

armenzg@armenzg-thinkpad:~/repos/ouija$ python -m timeit "open('/home/armenzg/.mozilla/seta/runnablejobs.json', 'r')"
100000 loops, best of 3: 8.09 usec per loop
```

tools/update_runnablejobs.py
armenzg: My apologies. I didn't realize that I had included the commented out code and the sys.exit()

jmaher:
yeah, that could be up for better code writing.  This is all new code, so
we are putting up what we think will work and go from there.

tools/update_runnablejobs.py
```diff
@@ -191,59 +260,88 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
-        testtype = job["testtype"]
-
-        found = False
-        found_id = None
+    # TODO: write test for this
+    # When the table is empty it means that we're starting the system for the first time
+    # and we're going to use different default values
+    map = {}
+    if not len(db_data) == 0:
+        priority, timeout = 1, 0
```
jmaher: I do not like assigning variables to a set of values, especially when the values are not the same (like None, or 0)

tools/update_runnablejobs.py
```diff
@@ -191,59 +260,88 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
-                # TODO: what if we have a race condition with two identical jobs
-                # verify the build system type is the same, or make it *
-                found = True
-                if row[7] != "*" and _buildsystem != row[7]:
+            key = tuple(row[1:4])
+            # This is guaranteed by a unique composite index for these 3 fields in models.py
+            assert key not in map,\
+                '"{}" should be a unique row and that is unexpected.'.format(key)
+            # (testtype, buildtype, platform)
+            map[key] = {'pk': row[0], 'build_system_type': row[7]}
```
jmaher: are we assuming that row[7] will have the '*' value already?

tools/update_runnablejobs.py
```diff
@@ -191,59 +260,88 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
+    else:
+        priority, timeout, expiration_date = 5, 5400, None
```
jmaher: please make unique lines for the assignment here.

tools/update_runnablejobs.py
```diff
@@ -191,59 +260,88 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
+
+    total_jobs, new_jobs, failed_changes, updated_jobs = len(data), 0, 0, 0
```
jmaher: and here as well, understanding len(data),0,0,0 is not easy to read

tools/update_runnablejobs.py
```diff
@@ -191,59 +260,88 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
 
-        # We have new jobs from runnablejobs to add to our master list
-        if not found:
-            _expired = None
-            if set_expired:
-                # set _expired = today + 14 days
-                # TODO: write test for it
-                _expired = "%s" % (datetime.datetime.now() + datetime.timedelta(days=14))
+                # Compare original build system to the new one
+                if _buildsystem != job['build_system_type']:
```
jmaher: you will only be in this if statement if you hit the condition in line 298, lets just have a single if statement here.

tools/update_runnablejobs.py
```diff
@@ -191,59 +260,88 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
+                    # XXX: Is this if check worth keeping? or should we join we the one above?
+                    #      Where we set _buildsystem to '*'
+                    # Update table with new buildsystem
+                    conn = engine.connect()
+                    statement = update(JobPriorities).where(
+                        JobPriorities.id == map[key]['pk_key']).values(buildsystem=_buildsystem)
+                    conn.execute(statement)
+                    LOG.info('Updated {} from {} to {}'.format(
+                        job['platform'], job['build_system_type'], _buildsystem
```
jmaher: this log doesn't indicate testtype,buildtype- the platform isn't that useful

tools/update_runnablejobs.py
```diff
@@ -191,59 +260,88 @@ def add_jobs_to_jobpriority(new_data=None, priority=1, timeout=0, set_expired=Fa
...
+                    expiration_date,
+                    _buildsystem
+                )
                 session.add(jobpriority)
                 session.commit()
-                added_jobs.append(job)
             except Exception as error:
                 session.rollback()
-                logging.warning(error)
+                LOG.warning(error)
```
jmaher: I would like to log specific job information here- it might help us see bad data or hunt down why a job is being scheduled differently.

database/models.py
```diff
@@ -98,6 +107,10 @@ class JobPriorities(MetaBase):
     expires = Column(DateTime)
     buildsystem = Column(String(64), nullable=False, index=True)
 
+    __table_args__ = (
+        UniqueConstraint('testtype', 'buildtype', 'platform', name='unique_job_priority'),
```
armenzg: Fixed in next push.

tools/failures.py
```diff
@@ -80,7 +89,7 @@ def communicate(failures, to_insert, total_detected, testmode, date):
 
 def reset_preseed():
     data = session.query(JobPriorities.expires, JobPriorities.id)\
-                  .filter(JobPriorities.expires != None).all()
+                  .filter(JobPriorities.expires is not None).all()
```
armenzg: It should be fine, however, since I have not tested this part of the code I will delay dealing with this at that point.

tools/failures.py
```diff
@@ -145,6 +153,13 @@ def parse_args(argv=None):
                         help="ending date for comparison."
                         )
 
+    parser.add_argument("--do-not-analyze",
+                        action="store_true",
+                        dest="do_not_analyze",
+                        help="With this option one can ignore root causes of revisions. \
+                              Specify the number of *extra* passes to be done."
+                        )
```
armenzg: Fixed in the next push. This allows to not call analyze_failures() down below.

tools/update_runnablejobs.py
```diff
@@ -1,98 +1,201 @@
...
+            map[key] = job['build_system_type']
+            sanitized_list.append(new_job)
+
+        elif job['build_system_type'] != map[key]:
+            previous_job = copy.deepcopy(new_job)
+            previous_job['build_system_type'] = map[key]
+            previous_index = sanitized_list.index(previous_job)
+            # This will make so we *replace* the previous build system type with '*'
+            # This also guarantees that we don't have duplicates
+            sanitized_list[previous_index]['build_system_type'] = '*'
```
armenzg: No, I don't. All I really care about is creating a sanitized list with the minimum data we need.
map is only used internally in this function to help me determine which jobs we have seen before and update the build_system_type if seen twice.

tools/update_runnablejobs.py
```diff
@@ -1,98 +1,201 @@
...
     try:
-        latest_task = retry(requests.get, args=(url % "mozilla-inbound", ),
-                            kwargs={'headers': {'accept-encoding': 'json'},
-                                    'verify': True}).json()
+        latest_task = retry(
+            requests.get,
+            args=(url % repo_name, ),
+            kwargs={'headers': {'accept-encoding': 'json'}, 'verify': True}
+        ).json()
+        task_id = latest_task['taskId']
```
armenzg: We were also doing the same previously.
We were checking for a timestamp which was being updated every time there is a new task.
The latest index always points to the latest gecko decision task.

What we could do is track if the size of the response has changed from the last time.

In any case, once we're on SETA we can probably query directly.

tools/update_runnablejobs.py
```diff
@@ -1,98 +1,201 @@
...
-            if len(data['results']) > 0:
-                data['meta'].update({'timetamp': new_timestamp})
-        except:
-            print "failed to get runnablejobs via %s" % url
-            print "We have something wrong with the runnable api, switch to the old runnablejobs.json now"
-            with open(get_rootdir() + '/runnablejobs.json', 'r') as data:
-                data = json.loads(data.read())
+        LOG.info("We're going to fetch new runnable jobs data.")
+        # We never store the output of runnable api but the minimal data we need
+        data = sanitized_data(query_the_runnablejobs(repo_name=repo_name, task_id=task_id))
```
armenzg: On Heroku we don't have to worry as the dynos are restarted every day and the files are not kept.
The files are also extremely little (303K).

armenzg: If we download the same data that the last gecko decision task produced (which is always the case minus when someone lands new jobs) we quickly skip over known data. I've tested the code multiple times and it is extremely fast.

I can optimize it even more if needed (which I don't believe it necessary atm) by discarding what was already seen in previous runs.

tools/update_runnablejobs.py
```diff
@@ -1,98 +1,201 @@
...
+        # Store the sanitized data to disk for inspection
+        with open(path, 'w') as f:
+            json.dump(data, f, indent=2, sort_keys=True)
+
         return data
 
 
-def parse_testtype(build_system_type, refdata, platform_option, job_type_name):
+# TODO we need add test for this function to make sure it works under any situation
+def update_jobs_priority_table():
```
armenzg: Fixed in next push.

tools/update_runnablejobs.py
```diff
@@ -133,55 +236,21 @@ def parse_testtype(build_system_type, refdata, platform_option, job_type_name):
...
-    return platform
+def valid_platform(platform):
+    # TODO: this is very hardcoded and prone to fall out of date
+    # XXX: What is the specific reason that these get excluded?
+    return platform not in [
+        'mulet-linux64', 'b2g-device-image', 'osx-10-7', 'osx-10-9', 'osx-10-11',
+        'windows8-32', 'android-4-2-armv7-api15', 'android-4-4-armv7-api15',
+        'android-5-0-armv8-api15', 'android-5-1-armv7-api15', 'android-6-0-armv8-api15',
+        'taskcluster-images', 'other', 'Win 6.3.9600 x86_64', 'windows7-64'
+    ]
```
armenzg: Fixed in the next push.

tools/update_runnablejobs.py
armenzg: I've seen this information on the logs.
We actually get to see the sql warning.

armenzg:
All comments are in.
I've address all changes.
If feedback is positive do not merge as I would like to squash the 2 last commits into the correct previous commits.

tools/update_runnablejobs.py
jmaher: got it, we can revisit this once fully migrated and we are into the final optimization phase.

tools/update_runnablejobs.py
jmaher: I am still worried about the case where we end up with an exception whilst calling sanitized_data(), it would be nice to know that we can always fall back to old data.

PR #159, opened 2016-10-07
make endpoint for taskcluster can works as expected

MikeLing:


MikeLing:
@jmaher r?

src/server.py
```diff
@@ -499,19 +503,19 @@ def run_seta_details_query():
                               JobPriorities.priority,
                               ).all()
 
-    # priority = 0; run all the jobs
-    if priority != 1 and priority != 5:
-        priority = 0
+        # priority = 0; run all the jobs
+        if priority != 1 and priority != 5:
+            priority = 0
```
jmaher: I still assert that we do not need priority 0- but I do see a scenario where we could have it.  This code block requires us to only support 1 or 5 for priority- lets leave this alone for now- we can deal with multiple priority support after this is live as a new feature.

PR #160, opened 2016-10-08
Fix the error in taskcluster endpoint

MikeLing:


src/server.py
```diff
@@ -408,7 +409,7 @@ def run_seta_details_query():
     buildbot = sanitize_bool(request.args.get("buildbot", 0))
     branch = sanitize_string(request.args.get("branch", ''))
     taskcluster = sanitize_bool(request.args.get("taskcluster", 0))
-    priority = int(sanitize_string(request.args.get("priority", 1)))
+    priority = int(sanitize_string(request.args.get("priority", 5)))
```
jmaher: this will default to low value jobs only- the goal of this function is to use the counts/time to determine which priority to use- ideally we will not have any priority variable passed in.

PR #165, opened 2016-10-18
Fix regression + add travis support + remove bad tests

armenzg:


tox.ini
```diff
@@ -23,4 +21,4 @@ commands =
 exclude = .tox
 show-source = True
 max-line-length=100
-ignore=D100,D101,D102,D103,D205,D400,D401
\ No newline at end of file
+ignore=D100,D101,D102,D103,D104,D105,D200,D201,D202,D203,D204,D205,D208,D210,D211,D300,D400,D401,D403,E123,F821,E225,F401,F841,E711,E501,D209,E303
```
jmaher: that is a LOT of ignores!

armenzg: There's a lot of warnings in the code. I did not want to have to fix them now. I can do a follow up.

jmaher:
ok, I am fine fixing it over time.

PR #167, opened 2016-10-18
Run update_runnablejobs.py on schedule

armenzg:
This will ensure that we have an up-to-date job priority table once a
day.

By adding this change, we can stop calling this script if
runnable_jobs.json is not on disk.

@MikeLing @jmaher what do you think?

I believe in this change we should also stop calling update_runnablejobs to prevent race conditions; what do you think?

### DIFF
--- a/schduler/clock.py
+++ b/schduler/clock.py
@@ -2,7 +2,11 @@ from rq import Queue
 from worker import conn
 import logging
 from apscheduler.schedulers.blocking import BlockingScheduler
-from trigger_jobs import trigger_migratedb, trigger_failures
+from trigger_jobs import (
+    trigger_failures,
+    trigger_migratedb,
+    trigger_update_job_priority_table,
+)
 
 
 logger = logging.info(__name__)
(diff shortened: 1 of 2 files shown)


MikeLing:
hmm, it seems like I haven't follow up for a long time. Why we need to run trigger_migratedb routinely? it seems like will call database_migration.py which is, if I remember correct, to migrate old databse on production-server to heroku.

schduler/clock.py
```diff
@@ -21,6 +25,11 @@ def timed_trigger_updatedb_sec():
     q.enqueue(trigger_migratedb)
 
 
+@sched.scheduled_job('cron', day_of_week='mon-sun', hour=12)
+def timed_trigger_updatedb_sec():
+    q.enqueue(trigger_migratedb)
```
jmaher: why are we adding a second trigger_migratedb here, I assume this was supposed to be  trigger_update_job_priority_table() ?

armenzg:
My apologies. I had few mistakes on that PR.

I'm only interested on updating the jobs priority table.

armenzg:
My lad push contains the fixes.

PR #199, opened 2016-12-11
makes ouija return all jobs for 90 minutes

MikeLing:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/ouija/199)

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -431,6 +431,10 @@ def run_seta_details_query():
         # can simplify the request url.
         taskcluster = 1
 
+        # we will return all jobs for every 90 minutes, so the return_all_jobs flag will been
+        # set to true if the time limit been reached.
+        return_all_jobs = False
+
         # We should return full job list as a fallback, if it's a request from
         # taskcluster and without head_rev or pushlog_id in there
         try:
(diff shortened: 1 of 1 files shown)


src/server.py
```diff
@@ -482,8 +486,8 @@ def run_seta_details_query():
                 datetime=time_of_now)
             conn.execute(statement)
 
-            # we must reset the delta after we update the datetime
-            delta = 0
```
MikeLing: we don't need to update delta to 0 due to it will been updated every push. And we could use it to see if we meet the 90 minutes limit in https://github.com/mozilla/ouija/pull/199

src/server.py
```diff
@@ -497,12 +501,17 @@ def run_seta_details_query():
         for d in query:
             # we only return that job if it hasn't reach the timeout limit. And the
             # timeout is zero means this job need always running.
-            if delta < d[4] or d[0] == 0:
+            if delta < d[4] or d[4] == 0:
```
MikeLing: d[0] is the job id, so I think d[4] (**timeout**) is more reasonable to me :)

MikeLing:
@jmaher r?

jmaher:
thank you for this MikeLing!

